PR #2, opened 2024-04-23
BAVL-66 intial changes to support running the API locally.

matt-crowson-moj:


src/main/resources/application-local.yml
```diff
@@ -0,0 +1,31 @@
+#
+# The "-local" profile is for running the API service locally against pre-existing
+# docker containers that are listening on a known port numbers for a postgres database (5432).
+#
+
+# Most values are inherited from - application.yml - except the following:
+spring:
+  datasource:
+    url: 'jdbc:postgresql://${DB_SERVER}/${DB_NAME}?sslmode=prefer'
```
matt-crowson-moj, line 9: We could possibly just hard code the values for the db and users here to save specifying them given they rarely change when running locally.

timharrison-moj: They're overridden in docker-compose and run-local scripts, so best left with env vars, I think.

timharrison-moj (approved):
LGTM.

PR #3, opened 2024-04-24
BAVL-67 adding config for HMPPS auth integration.

matt-crowson-moj:
This change includes the basics for getting integration with HMPPS auth using a common shared HMPPS Kotlin library.  

This takes away a lot of the boiler plate code we would otherwise have to include.  It also allows for consistency moving forwards should things change i.e. the share component can pick up this responsibility.

If push comes to shove we can use our own instead but it shouldn't come to that if we use the common one.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -8,7 +8,14 @@ configurations {
 }
 
 dependencies {
+  // Spring boot dependencies
   implementation("org.springframework.boot:spring-boot-starter-webflux")
+  implementation("uk.gov.justice.service.hmpps:hmpps-kotlin-spring-boot-starter:0.2.4")
+
+  // Test dependencies
+  testImplementation("io.jsonwebtoken:jjwt-impl:0.12.5")
+  testImplementation("io.jsonwebtoken:jjwt-jackson:0.12.5")
+  testImplementation("org.wiremock:wiremock-standalone:3.5.3")
 }
 
 kotlin {
--- a/src/main/resources/application-local.yml
+++ b/src/main/resources/application-local.yml
@@ -13,19 +13,14 @@ spring:
   jpa:
     show-sql: true
 
-  security:
-    oauth2:
-      resourceserver:
-        jwt:
-          public-key-location: classpath:local-public-key.pub
-
 system:
   client:
     id: ${SYSTEM_CLIENT_ID}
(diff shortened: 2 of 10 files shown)


build.gradle.kts
```diff
@@ -8,7 +8,14 @@ configurations {
 }
 
 dependencies {
+  // Spring boot dependencies
   implementation("org.springframework.boot:spring-boot-starter-webflux")
+  implementation("uk.gov.justice.service.hmpps:hmpps-kotlin-spring-boot-starter:0.2.4")
```
matt-crowson-moj, line 13: The key thing to be aware of here is this shared common component pulls in the necessary config for auth setup.  We don't need a dedicated class (it also gives us a entry point to providing SAR data when the time comes).

We will need one for our WebClient configs though.

src/main/resources/application-local.yml
```diff
@@ -13,19 +13,14 @@ spring:
...
     secret: ${SYSTEM_CLIENT_SECRET}
 
 # Container-provider local auth server
-hmpps:
-  auth:
-    url: https://sign-in-dev.hmpps.service.justice.gov.uk/auth
+api:
+  base:
+    url:
+      hmpps-auth: https://sign-in-dev.hmpps.service.justice.gov.uk/auth
```
matt-crowson-moj, line 25: We generally don't use a local container for auth anyway so setting to DEV is the logical call.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/NotFoundTest.kt
```diff
@@ -7,6 +7,7 @@ class NotFoundTest : IntegrationTestBase() {
   @Test
   fun `Resources that aren't found should return 404 - test of the exception handler`() {
     webTestClient.get().uri("/some-url-not-found")
+      .headers(setAuthorisation())
```
matt-crowson-moj, line 10: We now have auth so we need to provide a token in the header.

timharrison-moj (approved):
LGTM
We know this will probably need to call activities API to exchange booking info, so a step to get that relationship setup might be down the line a bit?

PR #4, opened 2024-04-25
BAVL-67 adding health check for HMPPS auth.

matt-crowson-moj:
Adds health check for recently added HMPPS auth integration.

Any services we rely on should be added as health checks going forwards.


timharrison-moj (approved):
LGTM

PR #5, opened 2024-04-26
BAVL-7: Flyway configured, initial data model and base data added.

timharrison-moj:
Initial data model and base data.
No location decoration or default CVP / PCVL links  yet.
Keeping simple for now so that migration in from current BVL data is easier.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -9,13 +9,26 @@ configurations {
 
 dependencies {
   // Spring boot dependencies
+  implementation("org.springframework.boot:spring-boot-starter-security")
   implementation("org.springframework.boot:spring-boot-starter-webflux")
   implementation("uk.gov.justice.service.hmpps:hmpps-kotlin-spring-boot-starter:0.2.4")
+  implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+
+  // Database dependencies
+  runtimeOnly("org.flywaydb:flyway-core")
+  runtimeOnly("org.postgresql:postgresql:42.7.1")
 
   // Test dependencies
   testImplementation("io.jsonwebtoken:jjwt-impl:0.12.5")
   testImplementation("io.jsonwebtoken:jjwt-jackson:0.12.5")
   testImplementation("org.wiremock:wiremock-standalone:3.5.3")
+  testImplementation("com.h2database:h2")
+  testImplementation("org.mockito:mockito-inline:5.2.0")
+  testImplementation("net.javacrumbs.json-unit:json-unit:3.2.2")
+  testImplementation("net.javacrumbs.json-unit:json-unit-assertj:3.2.2")
+  testImplementation("net.javacrumbs.json-unit:json-unit-json-path:3.2.2")
+  testImplementation("org.springframework.security:spring-security-test")
+  testImplementation("org.springframework.boot:spring-boot-starter-test")
 }
 
 kotlin {
--- a/src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
+++ b/src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
@@ -1,0 +1,328 @@
+--
+-- TABLES
+--
+
+---------------------------------------------------------------------------------------
+-- Contains coded reference values, used to constrain the values of lists/validation.
+-- e.g. Booking types, meeting types, hearing types, status etc..
+----------------------------------------------------------------------------------------
+
+CREATE TABLE reference_code
+(
+    reference_code_id   bigserial NOT NULL CONSTRAINT reference_code_pk PRIMARY KEY,
+    group_code          varchar(40) NOT NULL,  
+    code                varchar(40) NOT NULL,
+    description         varchar(100) NOT NULL,
+    created_by          varchar(100) NOT NULL,
+    created_time        timestamp NOT NULL,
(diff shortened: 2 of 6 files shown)


src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
```diff
@@ -0,0 +1,328 @@
...
+
+CREATE INDEX idx_court_code ON court(code);
+
+---------------------------------------------------------------------------------------
+-- The people involved in video link bookings at each court.
+-- This can be a single group email contact or a list of individual contacts.
+-- This list of contacts will be notified of bookings / changes for this court.
+----------------------------------------------------------------------------------------
+
+CREATE TABLE court_contact
```
matt-crowson-moj, line 52: The contact tables all look very similar.  Do we need 3 tables?  We could have a contact type?

I realise this means reference id gotchas.

src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
```diff
@@ -0,0 +1,328 @@
...
+    description  varchar(100) NOT NULL,
+    enabled      boolean NOT NULL,
+    notes        varchar(200),
+    created_by   varchar(100) NOT NULL,
+    created_time timestamp NOT NULL,
+    amended_by   varchar(100),
+    amended_time timestamp
+);
+
+CREATE INDEX idx_prison_code ON prison(code);
```
matt-crowson-moj, line 139: Add the UNIQUE here instead of on the `code` column? 🤷

src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
```diff
@@ -0,0 +1,328 @@
...
+-- court_id / hearing_type / probation_team_id / probation_meeting_type  
+--     - are nullable FK references
+----------------------------------------------------------------------------------------
+
+CREATE TABLE video_booking
+(
+    video_booking_id       bigserial NOT NULL CONSTRAINT video_booking_pk PRIMARY KEY,
+    booking_type           varchar(40) NOT NULL,  -- COURT, PROBATION
+    status_code            varchar(20) NOT NULL,  -- ACTIVE, COMPLETED, EXPIRED, CANCELLED
+    court_id               bigint REFERENCES court(court_id), -- Nullable
```
matt-crowson-moj, line 177: I am still not 100% sure on this. Relies on code enforcing the rules.

I wonder, is it possible to add constraints at the DB level e.g. one or the other must have a value?

build.gradle.kts
```diff
@@ -9,13 +9,26 @@ configurations {
...
+
+  // Database dependencies
+  runtimeOnly("org.flywaydb:flyway-core")
+  runtimeOnly("org.postgresql:postgresql:42.7.1")
 
   // Test dependencies
   testImplementation("io.jsonwebtoken:jjwt-impl:0.12.5")
   testImplementation("io.jsonwebtoken:jjwt-jackson:0.12.5")
   testImplementation("org.wiremock:wiremock-standalone:3.5.3")
+  testImplementation("com.h2database:h2")
```
matt-crowson-moj, line 25: OK for now to use H2 but in this project but ideally I'd really prefer to start using TestContainers.

If we do it from the outset shouldn't be too much bother.

build.gradle.kts
```diff
@@ -9,13 +9,26 @@ configurations {
 
 dependencies {
   // Spring boot dependencies
+  implementation("org.springframework.boot:spring-boot-starter-security")
```
matt-crowson-moj, line 12: Quite a few other dependencies introduced here.  Do we need them all?

Would prefer to add as and when we need them 🤷

matt-crowson-moj (dismissed):
LGTM

src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
timharrison-moj: I was thinking there might be more - police, parole - to come, and wondering whether the contact methods for them might change. I went separate just for clarity .. picturing 2-3 lists related to a booking of the contacts to keep informed about it. Good challenge, but one I did think about and went for separate (for now!) ..

src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
timharrison-moj: Yes. Will add now, thanks

src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
timharrison-moj: Don't know a way to constrain either/or - unless we have a pre-insert trigger and a procedure/function to check?

build.gradle.kts
timharrison-moj: Happy with a move to testcontainers so we get like-for-like postrgresql during tests, but ok for this merge I think. Will be a separate ticket to add TC pgsql.

build.gradle.kts
timharrison-moj: It was just in passing ... realised we'd need them soon, but can take out if you prefer?

matt-crowson-moj (approved):
LGTM

PR #6, opened 2024-04-26
BAVL-7 correcting the DB mode config value in the appliation yaml file.

matt-crowson-moj:


timharrison-moj (approved):
LGTM
Think we used to set this in an env var, so can alter locally and in K8s envs?

PR #7, opened 2024-04-26
BAVL-8: Setup DB properties from namespace secrets

timharrison-moj:
Introduce DB parameters from secrets.
Set DB SSL mode from environment.


matt-crowson-moj (approved):
LGTM

PR #10, opened 2024-04-29
BAVL-71 adding the necessary client for making calls to the prisoner search API.

matt-crowson-moj:
This change will allow our service to make calls to the external prisoner search API for checking the details of a prisoner related to video link bookings.

Note the original intention of this change was to generate the types for the prisoner search API.  However open API seemed to have an issue (bug) with generating it and I don't want to spend too long trying to see if there is a solution at this point in time.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -65,8 +65,8 @@ tasks {
 val configValues = mapOf(
   "dateLibrary" to "java8-localdatetime",
   "serializationLibrary" to "jackson",
-  "useBeanValidation" to "false",
-  "enumPropertyNaming" to "UPPERCASE",
+  "enumPropertyNaming" to "original",
+  "useSpringBoot3" to "true",
 )
 
 val buildDirectory: Directory = layout.buildDirectory.get()
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/prisonersearch/PrisonerSearchClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/prisonersearch/PrisonerSearchClient.kt
@@ -1,0 +1,23 @@
+package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.client.prisonersearch
+
+import org.springframework.web.reactive.function.client.WebClient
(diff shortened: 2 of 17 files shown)


build.gradle.kts
```diff
@@ -65,8 +65,8 @@ tasks {
 val configValues = mapOf(
   "dateLibrary" to "java8-localdatetime",
   "serializationLibrary" to "jackson",
-  "useBeanValidation" to "false",
-  "enumPropertyNaming" to "UPPERCASE",
+  "enumPropertyNaming" to "original",
+  "useSpringBoot3" to "true",
```
matt-crowson-moj, line 69: Whilst this makes no obvious impact we should be setting these here during generation of API's for spring boot 3 projects.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/prisonersearch/PrisonerSearchClient.kt
```diff
@@ -0,0 +1,23 @@
...
+      .get()
+      .uri("/prisoner/{prisonerNumber}", prisonerNumber)
+      .retrieve()
+      .bodyToMono(Prisoner::class.java)
+      .onErrorResume(WebClientResponseException.NotFound::class.java) { Mono.empty() }
+      .block()
+}
+
+// TODO add additional fields as and when needed e.g. ACTIVE/NOT ACTIVE in prison
+data class Prisoner(
```
matt-crowson-moj, line 20: As per comment in the description ideally I would have generated this type but open API generator was not playing nicely.

src/test/resources/application-test.yml
```diff
@@ -27,3 +27,4 @@ api:
     url:
       hmpps-auth: http://localhost:8090/auth
       locations-inside-prison: http://localhost:8091
+      prisoner-search: http://localhost:8092
```
timharrison-moj, line 30: Are we going with incremental ports for test, so
auth - 8090
locations - 8091
prisoner-search - 8092
xxxx - 8093
yyyy - 8094

matt-crowson-moj: That's what I thought to keep things simple.  Open to other suggestions if you have them.

PR #11, opened 2024-04-29
BAVL-71 adding the necessary client for making calls to the activities and appointments API.

matt-crowson-moj:
This change will allow our service to make calls to the external activities and appointments API for dealing with video link appointments between the two services.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/wiremock/ActivitiesAppointmentsApiMockServer.kt
```diff
@@ -0,0 +1,69 @@
...
+
+  fun stubGetAppointment(appointmentId: Long) {
+    stubFor(
+      WireMock.get("/appointments/$appointmentId")
+        .willReturn(
+          WireMock.aResponse()
+            .withHeader("Content-Type", "application/json")
+            .withBody(
+              mapper.writeValueAsString(
+                Appointment(
```
matt-crowson-moj, line 24: We can pull this out into a factory class as and when necessary.

build.gradle.kts
```diff
@@ -72,15 +72,24 @@ val configValues = mapOf(
 val buildDirectory: Directory = layout.buildDirectory.get()
 
 tasks.register("buildLocationsInsidePrisonApiModel", GenerateTask::class) {
-  generatorName.set("kotlin-spring")
+  generatorName.set("kotlin")
```
matt-crowson-moj, line 75: Changed as kotlin-spring was generating OffsetDateTime fields when what we actually want is LocalDateTime fields.

src/main/resources/application.yml
```diff
@@ -38,6 +38,12 @@ spring:
           jwk-set-uri: ${api.base.url.hmpps-auth}/.well-known/jwks.json
       client:
         registration:
+          activities-appointments:
+            provider: hmpps-auth
+            client-id: ${system.client.id}
+            client-secret: ${system.client.secret}
+            authorization-grant-type: client_credentials
+            scope: read
```
timharrison-moj, line 46: Probably will need to be read/write here - but fine for now.

timharrison-moj (approved):
LGTM.

src/main/resources/application.yml
matt-crowson-moj: Good spot.

timharrison-moj (approved):
LGTM

PR #13, opened 2024-05-01
BAVL-14 this change adds a simple implementation for the Gov Notify email service.

matt-crowson-moj:
This change provides a simple implementation for sending emails via Gov Notify.

The service will likely need further changes but hopefully this gives us a good initial footing.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/GovNotifyEmailService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/GovNotifyEmailService.kt
@@ -1,11 +1,59 @@
 package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service
 
+import org.slf4j.LoggerFactory
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.config.Email
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.config.EmailService
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.config.EmailTemplates
 import uk.gov.service.notify.NotificationClient
 
-internal class GovNotifyEmailService(client: NotificationClient, emailTemplates: EmailTemplates) : EmailService {
-  override fun sendEmail() {
-    TODO("Service is not yet implemented")
+class GovNotifyEmailService(
+  private val client: NotificationClient,
+  private val emailTemplates: EmailTemplates,
+) : EmailService {
+
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  override fun send(email: Email) {
+    when (email) {
+      is AppointmentCancelledCourtEmail -> send(email, emailTemplates.appointmentCancelledCourt)
+      is AppointmentCancelledPrisonEmail -> send(email, emailTemplates.appointmentCancelledPrison)
+      is AppointmentCancelledPrisonNoCourtEmail -> send(email, emailTemplates.appointmentCancelledPrisonNoCourt)
+      is OffenderReleasedCourtEmail -> send(email, emailTemplates.offenderReleasedCourt)
+      is OffenderReleasedPrisonEmail -> send(email, emailTemplates.offenderReleasedPrison)
+      is OffenderReleasedPrisonNoCourtEmail -> send(email, emailTemplates.offenderReleasedPrisonNoCourt)
+      is OffenderTransferredCourtEmail -> send(email, emailTemplates.offenderTransferredCourt)
+      is OffenderTransferredPrisonEmail -> send(email, emailTemplates.offenderTransferredPrison)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/GovNotifyEmailService.kt
```diff
@@ -1,11 +1,59 @@
...
+      is OffenderTransferredCourtEmail -> send(email, emailTemplates.offenderTransferredCourt)
+      is OffenderTransferredPrisonEmail -> send(email, emailTemplates.offenderTransferredPrison)
+      is OffenderTransferredPrisonNoCourtEmail -> send(email, emailTemplates.offenderTransferredPrisonNoCourt)
+      else -> throw RuntimeException("Unsupported email type ${email.javaClass.simpleName}.")
+    }
+  }
+
+  private fun send(email: Email, templateId: String) {
+    runCatching {
+      client.sendEmail(templateId, email.address, email.personalisation, null)
```
matt-crowson-moj, line 35: This will likely need tweaking but should suffice until we start using the service.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/GovNotifyEmailService.kt
```diff
@@ -1,11 +1,59 @@
...
+  private val emailTemplates: EmailTemplates,
+) : EmailService {
+
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  override fun send(email: Email) {
+    when (email) {
+      is AppointmentCancelledCourtEmail -> send(email, emailTemplates.appointmentCancelledCourt)
```
matt-crowson-moj, line 20: We could introduce an enum type that is shared between email impl and the email templates to allow us to pull out the template by type.  This `when` technically would become redundant e.g.

`send(email, emailTemplates.getTemplate(email.templateType)`

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/GovNotifyEmailService.kt
```diff
@@ -1,11 +1,59 @@
...
+    }
+  }
+
+  private fun send(email: Email, templateId: String) {
+    runCatching {
+      client.sendEmail(templateId, email.address, email.personalisation, null)
+    }
+      .onSuccess { log.info("EMAIL: sent ${email.javaClass.simpleName} email.") }
+      .onFailure { log.info("EMAIL: failed to send ${email.javaClass.simpleName} email.") }
+      .getOrThrow()
```
matt-crowson-moj, line 39: Something to think about may be in future.

If get an exception during email sending how should the service as whole behave?  Should the transaction fail for example?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/GovNotifyEmailService.kt
timharrison-moj: Yes, I think its a good idea to store the ID.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/GovNotifyEmailService.kt
timharrison-moj: I don't think it should fail the tx. Email notification issues should probably be alerted though. I think there might be some admin functionality we could do around resending notifications for a booking too?  But later..

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/GovNotifyEmailService.kt
timharrison-moj: I think the `when` is fine - clear what is happening.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/GovNotifyEmailService.kt
matt-crowson-moj: OK then this should not be propagating the exception (the `.getOrThrow()` bit)...

It should return a Result<IDENTIFIER> and the caller decides what action to take e.g. record the fact email as not sent successfully.

timharrison-moj (approved):
LGTM.
I think we should make the NOTIFY_API_KEY, and all the template IDs namespace secrets, pulled by values.yaml per environment. We can then work out how to avoid sending to real people - think the API key can be specific, so one for internal users (specified as a list in Notify) for DEV/PRE, and another for PROD which will send to anyone.

PR #14, opened 2024-05-02
BAVL-14 handle the sending of email errors gracefully, do not just throw error if send fails, let the caller decided on course of action.

matt-crowson-moj:
Handle email failures gracefully so the caller can decide on course of action if/when it happens.

The change also now returns the unique reference associated with an email send from gov notify so we can record it against the email notification in the notifications table.


timharrison-moj (approved):
LGTM

PR #15, opened 2024-05-02
BAVL-14 updates the notification table to include a reference to the associated gov notify notifications.

matt-crowson-moj:
The changes adds the notification id returned by gov notify to our own notifications table.

This will be useful down the line if we need to investigate any potential issues with notifications.  It also gives us the option to consider implementing a callback endpoint which gov notify can call to say whether some has actually been delivered or not.

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -8,6 +8,7 @@ generic-service:
     host: book-a-video-link-api-dev.prison.service.justice.gov.uk
 
   env:
+    SPRING_PROFILES_ACTIVE: dev
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.dev.json
     API_BASE_URL_ACTIVITIES_APPOINTMENTS: "https://activities-api-dev.prison.service.justice.gov.uk"
     API_BASE_URL_HMPPS_AUTH: "https://sign-in-dev.hmpps.service.justice.gov.uk/auth"
--- a/src/main/resources/application-dev.yml
+++ b/src/main/resources/application-dev.yml
@@ -1,0 +1,5 @@
+spring:
+  flyway:
+    # This should be removed when we think the DB is stable, and we should be moving onto migrations for changes.
+    clean-disabled: false
+    clean-on-validation-error: true
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -24,7 +24,7 @@ spring:
       ddl-auto: none
 
   flyway:
-    locations: classpath:/migrations/common,classpath:/migrations/prod\
+    locations: classpath:/migrations/common,classpath:/migrations/prod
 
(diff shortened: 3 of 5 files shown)


src/main/resources/application-dev.yml
```diff
@@ -0,0 +1,5 @@
+spring:
+  flyway:
+    # This should be removed when we think the DB is stable, and we should be moving onto migrations for changes.
```
matt-crowson-moj, line 3: @timharrison-moj 

I have done this so we can make tweaks to the baseline SQL until it we think it is fairly settled.

This avoids the need to write extra migrations in the early stages of the project.

We can always take it out if it becomes a problem.  The trade off here being we will loose any dev data whilst this is like this.

timharrison-moj: What is the effect of these?

src/main/resources/application.yml
```diff
@@ -24,7 +24,7 @@ spring:
       ddl-auto: none
 
   flyway:
-    locations: classpath:/migrations/common,classpath:/migrations/prod
+    locations: classpath:/migrations/common,classpath:/migrations/prod\
```
timharrison-moj, line 28: The slash on the end a typo?

matt-crowson-moj: darn it!!

helm_deploy/values-dev.yaml
```diff
@@ -8,6 +8,7 @@ generic-service:
     host: book-a-video-link-api-dev.prison.service.justice.gov.uk
 
   env:
+    SPRING_PROFILES_ACTIVE: dev
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.dev.json
```
timharrison-moj, line 12: Does it need this? The spring profile dev is usually for running locally, isn't it?

timharrison-moj (approved):
Couple of questions, but LGTM.

helm_deploy/values-dev.yaml
matt-crowson-moj: There is a local profile for running locally.

There is the dev profile added as part of this PR so we can override values the default application.yml file.

This is telling Spring to override the main profile with the dev profile file when deploying to dev.

src/main/resources/application-dev.yml
matt-crowson-moj: Without this we will have add (new) migrations for any changes to the base line DB.

It will mean we can loose what is in dev whilst this remains.  It would be nice not to have to add new migrations this early in development especially as things are likely to change.

The other alternative is drop the dev db when we make changes to the base line prior to deployment.

timharrison-moj: But with it - what happens to the existing table when it is changed? Does it drop/recreate?

timharrison-moj (approved):
LGTM

PR #16, opened 2024-05-03
BAVL-14 add gov notify api key environment var. Still need template IDs adding when we know them.

matt-crowson-moj:
This secret for the API key has been added to the dev k8s namespace.  I would suggest verifying this prior to approving the PR.


helm_deploy/hmpps-book-a-video-link-api/values.yaml
```diff
@@ -36,6 +37,8 @@ generic-service:
       DB_NAME: "database_name"
       DB_USER: "database_username"
       DB_PASS: "database_password"
+    gov-notify:
+      NOTIFY_API_KEY: "notify_api_key"
```
timharrison-moj, line 42: Thinking we'd add the trial key here (for now)? Then later - this can be overridden in PROD with another API key which can send to anyone?

matt-crowson-moj: Yarp

PR #21, opened 2024-05-10
BAVL-97 adds the columns/identifiers needed for courts and probation to the creation request.

matt-crowson-moj:
The change adds the necessary attributes to the create journey request to identify court and probation bookings.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/model/request/CreateVideoBookingRequest.kt
```diff
@@ -53,10 +104,6 @@ data class PrisonerDetails(
 )
 
 data class Appointment(
-  @field:NotNull(message = "The appointment type for the appointment is mandatory")
-  @Schema(description = "The appointment type", example = "TO BE DEFINED")
-  val type: AppointmentType?,
```
matt-crowson-moj, line 58: I don't think this is needed as we can derive this from the booking type.

timharrison-moj (approved):
LGTM

PR #23, opened 2024-05-13
BAVL-98 first simple integration test for the creation journey.

matt-crowson-moj:
The is the first end to end integration test which exercises the persistence of a court and probation booking to the database.

It does not create any appointments, this is next on the agenda.

I was thinking of not bothering with Controller dedicated tests as we really end up writing almost the same as an integration test anyway.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/resource/VideoLinkBookingIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/resource/VideoLinkBookingIntegrationTest.kt
@@ -1,0 +1,70 @@
+package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.integration.resource
+
+import org.junit.jupiter.api.Test
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.http.MediaType
+import org.springframework.test.web.reactive.server.WebTestClient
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/resource/VideoLinkBookingIntegrationTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.springframework.test.web.reactive.server.WebTestClient
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.helper.courtBookingRequest
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.helper.hasSize
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.helper.isEqualTo
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.helper.probationBookingRequest
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.integration.IntegrationTestBase
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.model.request.CreateVideoBookingRequest
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.repository.VideoBookingRepository
+
+class VideoLinkBookingIntegrationTest : IntegrationTestBase() {
```
matt-crowson-moj, line 15: As it currently stands this is seeding data from the baseline data in the resources folder.

This will need to change but for a quick starter which exercises the full end to end of the endpoint I think this is a reasonable compromise.

PR #24, opened 2024-05-14
BAVL-98 changes introduces and persists the appointment for a probation team appointment. Still requires further validation checks.

matt-crowson-moj:
This change introduces the PrisonAppointment entity and also creates one as part of the probation booking/appointment creation journey.  There are still validation needs in this area, e.g. checking of location keys.

Court appointments are not created in these changes.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/PrisonAppointment.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/PrisonAppointment.kt
@@ -1,0 +1,90 @@
+package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.entity
+
+import jakarta.persistence.Entity
+import jakarta.persistence.GeneratedValue
+import jakarta.persistence.GenerationType
+import jakarta.persistence.Id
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/PrisonAppointment.kt
```diff
@@ -0,0 +1,90 @@
...
+      endTime: LocalTime,
+      locationKey: String,
+      createdBy: String,
+    ) = PrisonAppointment(
+      videoBooking = videoBooking,
+      prisonCode = prisonCode,
+      prisonerNumber = prisonerNumber,
+      appointmentType = appointmentType,
+      appointmentDate = appointmentDate,
+      startTime = startTime.toMinutePrecision(),
```
matt-crowson-moj, line 84: We are not interested in seconds for times.

This will help with indexing as well as checking for overlapping date/times.

matt-crowson-moj:
> Code looks sound to me although I probably need to spend time getting a better understanding of business logic



> Code looks sound to me although I probably need to spend time getting a better understanding of business logic

There is still quite a bit more to come in.  Court appointment creation, validation etc.  Trying to keep PR's small.

PR #25, opened 2024-05-14
BAVL-98 changes to persist the appointments for a court hearing booking. Still requires further validation checks.

matt-crowson-moj:
This change now adds the creation of appointments for a court bookings.

There are further validation steps required but this gets the end to end flow in place.


timharrison-moj (approved):
LGTM

PR #42, opened 2024-05-28
BAVL-128 correct email contact to owner for new court email and update locations api to use new endpoint.

matt-crowson-moj:
This changes a call to the locations api to use a new post endpoint for locations so we can make one call instead of many calls for locations.

It also corrects the (initial) email contact type.  There are still more emails to be sent on creation of a booking.


timharrison-moj (approved):
LGTM

PR #44, opened 2024-05-29
BAVL-134: Add owner/creator contact details into the booking contacts service

timharrison-moj:
Also minor amendments:
* A court base data was incorrect.
* Add the column created_by_prison to the video booking, entity and model to show whether a court or probation booking was created by a prison (even though we do not have the process to do so as yet)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookingContactsService.kt
```diff
@@ -1,12 +1,55 @@
...
+            videoBookingId = videoBookingId,
+            contactType = ContactType.OWNER,
+            name = booking.amendedBy,
+            email = booking.amendedBy,
+          ),
+        )
+      }
+    }
+
+    return listOfContacts
```
matt-crowson-moj, line 53: Trivia but this will return a mutable list.  Just needs a `.toList()` on the end.

matt-crowson-moj:
I think you could possibly remove the mocking of the `BookingContactsService` in the `VideoLinkBookingIntegrationTest` as part of this change ...

matt-crowson-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookingContactsService.kt
timharrison-moj: I'll do it now.

timharrison-moj: Actually, I'll merge and do this as a follow up..

PR #47, opened 2024-05-31
BAVL-134 added checks to ensure we only send owner emails to verified email addresses only.

matt-crowson-moj:
Strictly speaking we should not need this (default call to users API should only return verified) but I have added as a precautionary measure given the change/impact is very small.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/manageusers/ManageUsersClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/manageusers/ManageUsersClient.kt
@@ -20,11 +20,11 @@ class ManageUsersClient(private val manageUsersApiWebClient: WebClient) {
   fun getUsersEmail(username: String): EmailAddressDto? =
     manageUsersApiWebClient
       .get()
-      .uri("/users/{username}/email", username)
+      .uri("/users/{username}/email?unverified=false", username)
       .retrieve()
       .bodyToMono(EmailAddressDto::class.java)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/manageusers/ManageUsersClient.kt
```diff
@@ -20,11 +20,11 @@ class ManageUsersClient(private val manageUsersApiWebClient: WebClient) {
   fun getUsersEmail(username: String): EmailAddressDto? =
     manageUsersApiWebClient
       .get()
-      .uri("/users/{username}/email", username)
+      .uri("/users/{username}/email?unverified=false", username)
```
matt-crowson-moj, line 23: This is the default anyway but added as a precautionary measure.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/manageusers/ManageUsersClient.kt
```diff
@@ -20,11 +20,11 @@ class ManageUsersClient(private val manageUsersApiWebClient: WebClient) {
...
       .retrieve()
       .bodyToMono(EmailAddressDto::class.java)
       .onErrorResume(WebClientResponseException.NotFound::class.java) { Mono.empty() }
-      .block()
+      .block()?.takeIf(EmailAddressDto::verified)
```
matt-crowson-moj, line 27: Same here added as a precautionary measure.

PR #48, opened 2024-05-31
BAVL-142 changes to prevent sumbission of invalid dates which would be otherwise be rounded as a result of default lenient dates.

matt-crowson-moj:
So I struggled to get a unit or integration test to confirm these changes work, I have tested this manually.

Perhaps have a quick read [here](https://makemeengr.com/uuuu-versus-yyyy-in-datetimeformatter-formatting-pattern-codes-in-java/) and [here](https://docs.oracle.com/javase/8/docs/api/java/time/format/DateTimeFormatter.html).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/config/HmppsBookAVideoLinkApiExceptionHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/config/HmppsBookAVideoLinkApiExceptionHandler.kt
@@ -8,6 +8,7 @@ import org.springframework.http.HttpStatus.BAD_REQUEST
 import org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR
 import org.springframework.http.HttpStatus.NOT_FOUND
 import org.springframework.http.ResponseEntity
+import org.springframework.http.converter.HttpMessageNotReadableException
 import org.springframework.web.bind.annotation.ExceptionHandler
 import org.springframework.web.bind.annotation.RestControllerAdvice
 import org.springframework.web.servlet.resource.NoResourceFoundException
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/model/request/CreateVideoBookingRequest.kt
```diff
@@ -137,7 +137,7 @@ data class Appointment(
   @field:NotNull(message = "The date for the appointment is mandatory")
   @field:Future(message = "The date for the appointment must be in the future")
   @Schema(description = "The future date for which the appointment will start", example = "2022-12-23")
-  @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "yyyy-MM-dd")
+  @JsonFormat(shape = JsonFormat.Shape.STRING, pattern = "uuuu-MM-dd")
```
matt-crowson-moj, line 140: This was the only way I could get it to accept the leniency configuration changes.

I struggled to find any real examples of why this would be.  `yyyy` represents an era date where as `uuuu`.  Given we are not recording BCE (before common era) dates this is fine.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/config/HmppsBookAVideoLinkApiExceptionHandler.kt
```diff
@@ -58,7 +59,7 @@ class HmppsBookAVideoLinkApiExceptionHandler {
       ),
     ).also { log.error("Unexpected exception", e) }
 
-  @ExceptionHandler(IllegalArgumentException::class)
+  @ExceptionHandler(IllegalArgumentException::class, HttpMessageNotReadableException::class)
```
matt-crowson-moj, line 62: So this deals with converting a 500 (for a bad date) to a 400 bad request.

timharrison-moj (approved):
LGTM

PR #53, opened 2024-06-05
BAVL-165 sorting of prisons by name and locations by description when returned via endpoints.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #67, opened 2024-06-17
BAVL-25 refactoring the events code, in particular around having a single abstract DomainEvent class to base events off.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/DomainEvents.kt
```diff
@@ -6,60 +6,46 @@ import java.time.LocalDateTime
...
+  val eventType = eventType.eventType
+  val description = eventType.description
+  val version: String = "1"
+  val occurredAt: LocalDateTime = LocalDateTime.now()
+
+  fun toEventType() = DomainEventType.valueOf(eventType)
 }
 
-data class AppointmentInformation(val appointmentId: Long) : AdditionalInformation
+interface AdditionalInformation
```
timharrison-moj, line 22: Empty interface? What value does this provide?

timharrison-moj: Just collecting the different event content classes into an interface, but nothing actually common between them?

timharrison-moj (approved):
LGTM. One question on empty interface.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/DomainEvents.kt
matt-crowson-moj: This isn't new, it was here before.  This interface is needed as it part of the message payload additionalInformation.  For each event type it would be different.  At the moment we just have the entity ID's in here but there is probably more to go in here but we just haven't got it yet e.g. base url which would be a link to the resource in question.

PR #68, opened 2024-06-17
BAVL-177: Exclude existing booking while carrying out availability check

smcveigh941:
Exclude a specific booking from the availability check. For example, exclude the appointment which you are amending.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/AvailabilityServiceTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/AvailabilityServiceTest.kt
@@ -218,4 +218,54 @@ class AvailabilityServiceTest {
       assertThat(alternatives).hasSize(0)
     }
   }
+
+  @Test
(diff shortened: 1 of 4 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/AvailabilityServiceTest.kt
```diff
@@ -218,4 +218,54 @@ class AvailabilityServiceTest {
...
+      mainAppointment = LocationAndInterval(
+        prisonLocKey = room1,
+        interval = Interval(start = LocalTime.of(10, 0), end = LocalTime.of(11, 0)),
+      ),
+      vlbIdToExclude = 5L,
+    )
+
+    val response = service.checkAvailability(request)
+
+    // We should not exclude appointments with videoBookingId == 2L (conflict exists)
```
matt-crowson-moj, line 264: Not sure if I an reading this wrong but this says `2L` in the comment but looks like `5L` in the request?

matt-crowson-moj (approved):
LGTM

PR #69, opened 2024-06-17
BAVL-25 adding config and wiring for SQS.

matt-crowson-moj:
Not 100% this is entirely correct.  May need to revert the PR if the service fails to start after merging into main.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/InboundEventsListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/InboundEventsListener.kt
@@ -4,12 +4,17 @@ import com.fasterxml.jackson.databind.ObjectMapper
 import com.fasterxml.jackson.databind.PropertyNamingStrategies
 import com.fasterxml.jackson.databind.annotation.JsonNaming
 import com.fasterxml.jackson.module.kotlin.readValue
+import io.awspring.cloud.sqs.annotation.SqsListener
+import io.opentelemetry.api.trace.SpanKind
+import io.opentelemetry.instrumentation.annotations.WithSpan
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import org.springframework.context.annotation.Profile
 import org.springframework.stereotype.Component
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/InboundEventsListener.kt
```diff
@@ -20,6 +25,8 @@ class InboundEventsListener(
     private val log: Logger = LoggerFactory.getLogger(this::class.java)
   }
 
+  @SqsListener("bvls", factory = "hmppsQueueContainerFactoryProxy")
+  @WithSpan(value = "Digital-Prison-Services-book_a_video_link_queue", kind = SpanKind.SERVER)
   fun onMessage(rawMessage: String) {
     if (features.isEnabled(Feature.SNS_ENABLED)) {
```
matt-crowson-moj, line 31: Messages are not sent of received as the feature is disabled by default.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/InboundEventsListener.kt
```diff
@@ -4,12 +4,17 @@ import com.fasterxml.jackson.databind.ObjectMapper
...
+import io.opentelemetry.api.trace.SpanKind
+import io.opentelemetry.instrumentation.annotations.WithSpan
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
+import org.springframework.context.annotation.Profile
 import org.springframework.stereotype.Component
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.config.Feature
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.config.FeatureSwitches
 
+@Profile("!test && !local")
```
matt-crowson-moj, line 17: Unless we have local stack running this needs disabling.  Running tests locally will fail and the service will not start locally either.

timharrison-moj (approved):
LGTM. Give it a try, can always revert if doesn't start up.

PR #82, opened 2024-06-18
BAVL-25 switching send/receive events on in dev.

matt-crowson-moj:
Flicks the switch to **ON** for creating bookings (appointments) in external services.

Gone a bit heavier on the logging just so we can see it though better in early stages.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/InboundEventsListener.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/InboundEventsListener.kt
@@ -26,8 +26,10 @@ class InboundEventsListener(
   }
 
   @SqsListener("bvls", factory = "hmppsQueueContainerFactoryProxy")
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/InboundEventsListener.kt
```diff
@@ -26,8 +26,10 @@ class InboundEventsListener(
   }
 
   @SqsListener("bvls", factory = "hmppsQueueContainerFactoryProxy")
-  @WithSpan(value = "Digital-Prison-Services-book_a_video_link_queue", kind = SpanKind.SERVER)
+  @WithSpan(value = "farsight-devs-book_a_video_link_queue", kind = SpanKind.SERVER)
   fun onMessage(rawMessage: String) {
+    log.info("LISTENER: raw message $rawMessage")
```
matt-crowson-moj, line 31: Will probably look at taking this out.  Added for initial running in case of any issues/problems.

PR #83, opened 2024-06-18
BAVL-25 fixing very subtle bug with check event value and not type.

matt-crowson-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/InboundEventsListenerTest.kt
```diff
@@ -80,6 +80,6 @@ class InboundEventsListenerTest {
       "Notification",
       mapper.writeValueAsString(event),
       UUID.randomUUID().toString(),
-      MessageAttributes(EventType("String", event.eventType)),
+      MessageAttributes(EventType(Type = "String", Value = event.eventType)),
```
matt-crowson-moj, line 83: This one slipped in cos both are strings and not using named params 🤦

timharrison-moj (approved):
LGTM

PR #88, opened 2024-06-20
BAVL-199: Create the booking and appointment history tables

timharrison-moj:
Just adds the new tables ahead of populating them in a following PR.


matt-crowson-moj (approved):
LGTM

PR #91, opened 2024-06-21
Task/bavl 25 external api calls for get appointments

matt-crowson-moj:
I have been able to test this with calling prison-api and again we are in a position where we have to override the return type as fields generated as mandatory are not being populated when the api is called.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/activitiesappointments/ActivitiesAppointmentsClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/activitiesappointments/ActivitiesAppointmentsClient.kt
@@ -1,12 +1,11 @@
 package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.client.activitiesappointments
 
-import org.slf4j.Logger
-import org.slf4j.LoggerFactory
+import org.springframework.core.ParameterizedTypeReference
 import org.springframework.stereotype.Component
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.WebClientResponseException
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/activitiesappointments/ActivitiesAppointmentsClient.kt
```diff
@@ -70,4 +58,21 @@ class ActivitiesAppointmentsClient(private val activitiesAppointmentsApiWebClien
       .retrieve()
       .bodyToMono(AppointmentSeries::class.java)
       .block()
+
+  fun getPrisonersAppointments(prisonCode: String, prisonerNumber: String, onDate: LocalDate): List<AppointmentSearchRequest> =
+    activitiesAppointmentsApiWebClient.post()
```
timharrison-moj, line 63: This RETURNS an AppointmentSearchRequest??  Not a Response object?

timharrison-moj (dismissed):
LGTM - just a question on response types for appointments.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/activitiesappointments/ActivitiesAppointmentsClient.kt
matt-crowson-moj: 🤔 good spot will check

matt-crowson-moj: Done

timharrison-moj (approved):
LGTM

PR #95, opened 2024-06-25
BAVL-213 intial work in progress commit to support cancellation of video bookings.

matt-crowson-moj:
This should be mergable. It is missing emitting of cancellation event and sending of cancellation emails.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
@@ -1,14 +1,22 @@
 package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.entity
 
+import jakarta.persistence.CascadeType
 import jakarta.persistence.Entity
+import jakarta.persistence.EnumType
+import jakarta.persistence.Enumerated
+import jakarta.persistence.FetchType
 import jakarta.persistence.GeneratedValue
 import jakarta.persistence.GenerationType
 import jakarta.persistence.Id
 import jakarta.persistence.JoinColumn
+import jakarta.persistence.OneToMany
 import jakarta.persistence.OneToOne
 import jakarta.persistence.Table
 import org.hibernate.Hibernate
+import java.time.LocalDate
 import java.time.LocalDateTime
+import java.time.LocalDateTime.now
+import java.time.LocalTime
 
 @Entity
 @Table(name = "video_booking")
@@ -18,8 +26,6 @@ class VideoBooking private constructor(
   val videoBookingId: Long = 0,
 
   val bookingType: String,
-
-  var statusCode: String = "ACTIVE",
 
   @OneToOne
   @JoinColumn(name = "court_id")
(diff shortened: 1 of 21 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
```diff
@@ -41,15 +47,58 @@ class VideoBooking private constructor(
 
   val createdBy: String,
 
-  val createdTime: LocalDateTime = LocalDateTime.now(),
+  val createdTime: LocalDateTime = now(),
 ) {
 
+  @OneToMany(mappedBy = "videoBooking", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
+  private val prisonAppointments: MutableList<PrisonAppointment> = mutableListOf()
```
matt-crowson-moj, line 54: I have been moving/thinking towards managing the addition of appointments to the booking.  We can still have logic in the service but having these here helps with the cancellation journey.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
```diff
@@ -41,15 +47,58 @@ class VideoBooking private constructor(
...
+  var statusCode: StatusCode = StatusCode.ACTIVE
+    private set
+
   var amendedBy: String? = null
 
   var amendedTime: LocalDateTime? = null
 
   fun isCourtBooking() = bookingType == "COURT"
 
+  fun addAppointment(
```
matt-crowson-moj, line 66: I have kept this fairly basic in terms of validation (none at the mo) at the moment but we can enrich if we feel the need e.g. prevent addition of duplicate appointments, appointments must be in future etc.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/handlers/VideoBookingCancelledEventHandler.kt
```diff
@@ -0,0 +1,33 @@
...
+
+  companion object {
+    private val log = LoggerFactory.getLogger(this::class.java)
+  }
+
+  override fun handle(event: VideoBookingCancelledEvent) {
+    videoBookingRepository.findById(event.additionalInformation.videoBookingId).ifPresentOrElse(
+      { booking ->
+        prisonAppointmentRepository.findByVideoBooking(booking).forEach {
+          // TODO - raise delete event or just call the manage external appointments service directly here?
```
matt-crowson-moj, line 23: We have a choice here to spawn off separate events to deal with removal of each appointment from external services.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
```diff
@@ -41,15 +47,60 @@ class VideoBooking private constructor(
...
+    )
+  }
+
+  fun cancel(cancelledBy: String) =
+    apply {
+      require(statusCode != StatusCode.CANCELLED) {
+        "Video booking $videoBookingId is already cancelled"
+      }
+
+      require(prisonAppointments.all { it.isStartsAfter(now()) }) { "Video booking $videoBookingId cannot be cancelled" }
```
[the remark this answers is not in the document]
matt-crowson-moj: Can do but I just wanted to run it by you both before I go making further changes.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
```diff
@@ -41,15 +47,60 @@ class VideoBooking private constructor(
 
   val createdBy: String,
 
-  val createdTime: LocalDateTime = LocalDateTime.now(),
+  val createdTime: LocalDateTime = now(),
 ) {
 
+  @OneToMany(mappedBy = "videoBooking", fetch = FetchType.LAZY, cascade = [CascadeType.ALL], orphanRemoval = true)
```
matt-crowson-moj, line 53: Perhaps this doesn't need to be LAZY either, there will never be that many on a single booking.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
timharrison-moj: Is it the thin edge of the wedge which takes us down the road of the big entity graph, which turned into a big issue in A&A? We currently deal with prison appointments separately to the booking - not linked entities so won't this require a reasonably large refactor? When we add booking history rows there is another potential list of related entities in a many to one with the booking? I would rather every pull of a booking did not bring back its entire history. Generally, I think I prefer views to get the data we need, and for updates / cancellations to be specific by ID - but I do see the logic for appointments and how much clearer the cancel code is. What we don't want is the UI code walking through large entity graphs though. 50/50 for me...

timharrison-moj: Think its ok ... provided we link entities very carefully, and only where we know there is no potential for growing data sets - like history rows.

matt-crowson-moj: And I think this is fine for CRUD operations which are ID based.  Retrieval is via views.

Not doing this makes the service layer overly complicated e.g. more dependencies meaning multiple repos to create a booking when it can be done with one.  We know what we did in A and A so I think this risk can be avoided.

We were also dealing with much larger data sets in A & A.

timharrison-moj (approved):
LGTM. Despite reservations over a growing entity graph.

timharrison-moj (approved):
LGTM

PR #100, opened 2024-06-27
BAVL-199: History fix when amending video link bookings.

timharrison-moj:
This was tested locally for create, multiple amendments, changing comments and cancelling.


matt-crowson-moj (approved):
LGTM

PR #101, opened 2024-06-27
BAVL-213 fixing mapping of inbound cancellation event. Mapped to incorrect event type.

matt-crowson-moj:
Added test coverage for bug.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/DomainEvents.kt
```diff
@@ -60,7 +60,7 @@ enum class DomainEventType(val eventType: String, val description: String) {
     "A video booking has been cancelled in the book a video link service",
   ) {
     override fun toInboundEvent(mapper: ObjectMapper, message: String) =
-      mapper.readValue<VideoBookingCreatedEvent>(message)
+      mapper.readValue<VideoBookingCancelledEvent>(message)
```
matt-crowson-moj, line 63: Doh 🤦

PR #103, opened 2024-06-27
BAVL-213 adding temporary logging for the cancellation of appointments in external services.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #111, opened 2024-07-01
BAVL-213 when cancelling (deleting) appointments in prison API we do not want the event in prison-api to be propagated.  This is the end of the road.

matt-crowson-moj:
This is mirroring the behavior in whereabouts-api.  We already do this for appointment creation in prison-api.


timharrison-moj (approved):
LGTM. 
I remember seeing this when looking at whereabouts-api, good spot.

PR #112, opened 2024-07-02
BAVL-235 adding support for processing of prisoner released events to handle potential cancellation of future bookings.

matt-crowson-moj:
The intention is to add localstack integration test coverage in a separate PR.

We also need a cloud platform change to take place to listen for the release event.  Until that happens the code for release events will not fire.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/InboundEventsListener.kt
```diff
@@ -28,8 +28,6 @@ class InboundEventsListener(
   @SqsListener("bvls", factory = "hmppsQueueContainerFactoryProxy")
   @WithSpan(value = "farsight-devs-book_a_video_link_queue", kind = SpanKind.SERVER)
   fun onMessage(rawMessage: String) {
-    log.info("LISTENER: raw message $rawMessage")
```
matt-crowson-moj, line 31: Taking out.  This will start to get quite noisy when we start listening for prisoner release events.

PR #113, opened 2024-07-02
BAVL-235 adding support for processing of prisoner released events to handle potential cancellation of future bookings.

matt-crowson-moj:
The intention is to add localstack integration test coverage in a separate PR.

We also need a cloud platform change to take place to listen for the release event. Until that happens the code for release events will not fire.


timharrison-moj (approved):
LGTM

PR #118, opened 2024-07-04
BAVL-228: Add additional tests around manage external appointments for amended bookings

timharrison-moj:


matt-crowson-moj (approved):
LGTM

PR #121, opened 2024-07-08
BAVL-235 fixing offset datetime issue spotted during testing with real release event in dev.

matt-crowson-moj:
The occurredAt is a an offset date time which we were not catering for.  This showed up a deserialisation issue when testing the release of a prisoner in dev with some video bookings.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/DomainEvents.kt
```diff
@@ -12,7 +13,7 @@ abstract class DomainEvent<T : AdditionalInformation>(
   val eventType = eventType.eventType
   val description = eventType.description
   val version: String = "1"
-  val occurredAt: LocalDateTime = LocalDateTime.now()
+  val occurredAt: String = LocalDateTime.now().toOffsetString()
```
matt-crowson-moj, line 16: Having this as a string is fine, we don't actually use it on receipt of events.  Having it as a local date time highlighted it was wrong anyway.

timharrison-moj (approved):
LGTM

PR #125, opened 2024-07-09
BAVL-243: Fix for emails being sent to the Owner when any action was being done to a booking

smcveigh941:
The work required in this area is not done, this is only part of everything which is required. Still missing, for example, is the email which a court should get when a prison user creates a booking


matt-crowson-moj (approved):
LGTM

PR #129, opened 2024-07-11
BAVL-246 changes to allow new bookings to be at the same date and time and location as cancelled ones, this is allowed.

matt-crowson-moj:
So this changes fixes the initial problem.  However we would potentially need different behavior if it were created by a prison user, as in we don't care about clashes at all when done by a prison user.


src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
```diff
@@ -223,7 +223,7 @@ CREATE INDEX idx_prison_appointment_loc_key ON prison_appointment(prison_loc_key
 CREATE INDEX idx_prison_appointment_date ON prison_appointment(appointment_date);
 CREATE INDEX idx_prison_appointment_start_time ON prison_appointment(start_time);
 CREATE INDEX idx_prison_appointment_end_time ON prison_appointment(end_time);
-CREATE UNIQUE INDEX idx_prison_appointment_loc_date_start_end on prison_appointment(prison_loc_key, appointment_date, start_time, end_time);
+CREATE INDEX idx_prison_appointment_loc_date_start_end on prison_appointment(prison_loc_key, appointment_date, start_time, end_time);
```
matt-crowson-moj, line 226: This had to be relaxed due to cancelled and active being the same date, times and locations.

timharrison-moj (approved):
LGTM

PR #141, opened 2024-07-25
BAVL-266 need to drop view if changing columns names and/or types.

matt-crowson-moj:


src/main/resources/migrations/common/R__v_video_booking_event.sql
```diff
@@ -1,3 +1,5 @@
+DROP VIEW IF EXISTS v_video_booking_event;
+
 CREATE OR REPLACE VIEW v_video_booking_event
```
matt-crowson-moj, line 3: probably don't need `replace` but left in for now.

PR #144, opened 2024-07-29
BAVL-266 adding location caching and capping the max number of days of data to 1 year.

matt-crowson-moj:
Changes to cache locations primarily for CSV extracts.

Also adding cap to prevent requesting more that 1 years worth of data.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/CsvDataExtractionService.kt
```diff
@@ -66,6 +75,12 @@ class CsvDataExtractionService(
       .writeValues(csvOutputStream.bufferedWriter())
       .use { writer -> probationEvents.forEach(writer::write) }
   }
+
+  private fun checkDaysBetweenDoesNotExceedAYear(fromDate: LocalDate, toDate: LocalDate) {
+    require(DAYS.between(fromDate, toDate) <= 365) {
```
matt-crowson-moj, line 80: Could set to 366 to cater for leap years?

PR #150, opened 2024-07-30
BAVL-266 requesting more than a year of data was returning a 500 error when it should be a 400.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/resource/CsvDataExtractionController.kt
```diff
@@ -120,19 +90,20 @@ class CsvDataExtractionController(private val service: CsvDataExtractionService)
...
-        "attachment; filename=video-links-by-probation-booking-date-from-${startDate.toIsoDate()}-for-$days-days.csv",
-      )
+    streamedCsvResponse(
+      startDate,
+      days,
+      "video-links-by-probation-booking-date-from-${startDate.toIsoDate()}-for-$days-days.csv",
+    ) { os -> service.probationBookingsByBookingDateToCsv(fromDate = startDate, toDate = startDate.plusDays(days), os) }
+
+  private fun streamedCsvResponse(startDate: LocalDate, days: Long, filename: String, block: (outputStream: OutputStream) -> Unit): ResponseEntity<StreamingResponseBody> {
+    require(ChronoUnit.DAYS.between(startDate, startDate.plusDays(days)) <= 365) {
```
matt-crowson-moj, line 100: Brought check into here as 400 bad request was not being thrown, 500 internal error was being raised when the same check is being carried out in the service layer.

The controller advice is not able to interpret it as it is was being treated as text and not JSON due to the set up of the ResponseEntity below.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/resource/CsvDataExtractionController.kt
```diff
@@ -15,14 +15,16 @@ import org.springframework.web.bind.annotation.RestController
...
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.common.toIsoDate
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service.CsvDataExtractionService
+import java.io.OutputStream
 import java.time.LocalDate
+import java.time.temporal.ChronoUnit
 
 @Tag(name = "CSV Data Extraction Controller")
 @RestController
-@RequestMapping(value = ["download-csv"])
+@RequestMapping(value = ["download-csv"], produces = ["text/csv", MediaType.APPLICATION_JSON_VALUE])
```
matt-crowson-moj, line 24: Adding JSON here allows for the controller advice to handle the exception (if thrown).

PR #167, opened 2024-08-14
BAVL-280 restrict access to create bookings for prison users enforced by their active caseload.

matt-crowson-moj:
Apologies this PR ballooned a little with some refactoring around use of users in the unit tests.

Essentially the core change prevents prison users from creating bookings not in their current active caseload.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/CreateVideoBookingService.kt
```diff
@@ -39,6 +40,8 @@ class CreateVideoBookingService(
     }
 
   private fun createCourt(request: CreateVideoBookingRequest, createdBy: User): Pair<VideoBooking, Prisoner> {
+    checkCaseLoadAccess(request.prisoner().prisonCode!!)
```
matt-crowson-moj, line 43: Only needed for courts bookings on create as prison users cannot create probation bookings.

PR #169, opened 2024-08-15
BAVL-280 restrict access to cancel bookings for prison users enforced by their active caseload.

matt-crowson-moj:
The cancel scenario showed a floor with earlier approach to caseload checks.  There won't always be a user in the request attributes e.g. service User.  However we can use the User passed into the service calls directly as they will always be available to be passed through.  The knock on effect of this is actually simpler code, so all good.

Note, search will is also impacted by this change, a prison user can only search for bookings in their active caseload.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/security/CaseloadAccess.kt
```diff
@@ -1,27 +1,14 @@
...
-import org.springframework.web.context.request.ServletRequestAttributes
-import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.config.getBvlsRequestContext
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service.User
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service.UserType
 
 /**
  * Caseload access checks are only relevant for prison users. If the current user is not a prison user then it will just
  * be ignored.
  */
-fun checkCaseLoadAccess(prisonCode: String) {
```
matt-crowson-moj, line 12: The cancel scenario showed a floor with this approach.  There won't always be a user in the request attributes e.g. service User.  However we can use the User passed into the service calls directly as they will always be available to be passed through.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/config/HmppsBookAVideoLinkApiExceptionHandler.kt
```diff
@@ -71,10 +72,10 @@ class HmppsBookAVideoLinkApiExceptionHandler : ResponseEntityExceptionHandler()
   fun handleCaseLoadAccessException(e: CaseloadAccessException): ResponseEntity<ErrorResponse> {
     log.info("Case load access exception: {}", e.message)
     return ResponseEntity
-      .status(NOT_FOUND)
+      .status(FORBIDDEN)
```
matt-crowson-moj, line 75: 404 NOT_FOUND only works for GET scenario so changed to 403 FORBIDDEN to cater for all other scenarios.

PR #172, opened 2024-08-16
BAVL-307 OWASP security updates.

matt-crowson-moj:
Update bumps the Circle CI ORB, JARs and Gradle wrapper.

I have tested this by running the security job on the pipeline and it goes green.


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM --platform=$BUILDPLATFORM eclipse-temurin:21-jre-jammy AS builder
+FROM --platform=$BUILDPLATFORM eclipse-temurin:21-jdk-jammy AS builder
```
matt-crowson-moj, line 1: So there has been a Slack thread on this but there was no real conclusion as to why this change is needed.

Either way this is just for building it and not the actual running i.e. still using JRE for running.

PR #173, opened 2024-08-19
BAVL-308 reason field size was being exceeded so reduced the length of the reasons provided.

matt-crowson-moj:
Rather than change DB just changed the actual comments going in to the notification table.  Personally I think this is fine with the reduced data.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookingFacade.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookingFacade.kt
@@ -129,16 +129,6 @@ class BookingFacade(
   }
 
   private fun sendEmailAndSaveNotification(email: Email, booking: VideoBooking, action: BookingAction) {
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookingFacade.kt
```diff
@@ -129,28 +129,18 @@ class BookingFacade(
...
 
   private fun sendEmailAndSaveNotification(email: Email, booking: VideoBooking, action: BookingAction) {
-    val bookingType = if (booking.isCourtBooking()) "court" else "probation"
-
-    val reason = when (action) {
-      BookingAction.CREATE -> "New $bookingType booking"
-      BookingAction.AMEND -> "Amended $bookingType booking"
-      BookingAction.CANCEL -> "Cancelled $bookingType booking"
-      BookingAction.RELEASED -> "Cancelled $bookingType booking due to release"
-      BookingAction.TRANSFERRED -> "Cancelled $bookingType booking due to transfer"
```
matt-crowson-moj, line 139: This was too long for probation transfers.

PR #175, opened 2024-08-20
BAVL-310 this is the first change towards suppoting the checking of a users ability to access court and probation bookings.

matt-crowson-moj:
This change extends the UserService to check an external users groups in terms of accessing court and probation bookings.

Further changes will be needed to excercise this.  Doing in parts so can rollback if any side effects of the change become apparent when merged/deployed in dev.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/UserService.kt
```diff
@@ -29,23 +32,56 @@ class UserService(private val manageUsersClient: ManageUsersClient) {
...
-        userType = userType,
-        name = userDetails.name,
-        email = if (username.isEmail()) username.lowercase() else manageUsersClient.getUsersEmail(username)?.email?.lowercase(),
-        activeCaseLoadId = userDetails.activeCaseLoadId?.takeIf { userType == UserType.PRISON },
-      )
     }
 }
 
-data class User(val username: String, private val userType: UserType, val name: String, val email: String? = null, val activeCaseLoadId: String? = null) {
+data class User(
```
matt-crowson-moj, line 66: This is feeling a little more complicated as a result in terms of permutations. 

Could pull each out into there our class/type? This may have ramifications elsewhere if we did this.

PR #181, opened 2024-08-23
BAVL-317 prevent changing of court and probation team, attributes are not updatable once created.

matt-crowson-moj:
Removing the court code and probation team attributes from the amend journey.

Because the attributes were marked as optional on the amend request object, removal of them should not impact the UI's calling this endpoint 🤞


.kotlin/sessions/kotlin-compiler-13933610130012879093.salive
matt-crowson-moj, line 1: This (temporary) file was committed in error.

timharrison-moj (approved):
LGTM

PR #186, opened 2024-09-09
BAVL-327 search should only find active bookings, cancelled bookings should be ignored during search.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/repository/VideoAppointmentRepository.kt
```diff
@@ -24,7 +24,18 @@ interface VideoAppointmentRepository : ReadOnlyRepository<VideoAppointment, Long
     forLocationKeys: List<String>,
   ): List<VideoAppointment>
 
-  fun findByPrisonerNumberAndAppointmentDateAndPrisonLocKeyAndStartTimeAndEndTime(
```
matt-crowson-moj, line 27: I could have just added the status code as part of the search but I really don't like these long function names albeit Spring generates the query for you.

PR #189, opened 2024-09-16
BAVL-343: Upgrade HMPPS SQS library to v4.4.2

smcveigh941:


matt-crowson-moj (approved):
LGTM.

Approved though not sure if there is anything else needs doing with this being a major version bump.

PR #192, opened 2024-09-17
BAVL-342 ensuring we keep created at and amended at inline with historic (migrated) bookings.

matt-crowson-moj:
These changes ensure not only do we keep the created at/by inline with the migrated records but also the amended at/by inline with the migrated bookings.

If an old booking was created a 2 months ago and amended one month ago then the new migrated booking should have the same values.  This ensure we keeps an accurate historic record on migrated records.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/migration/MigrationClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/migration/MigrationClient.kt
@@ -106,6 +106,15 @@ data class VideoBookingEvent(
   val postEndTime: LocalDateTime?,
 )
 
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/migration/MigrationClient.kt
```diff
@@ -67,106 +67,80 @@ class MigrationClient(
     .block()
 }
 
-// TODO temporary placeholder type until we the complated response object from the whereabouts-api
-@Deprecated(message = "temporary placeholder type until we the complated response object from the whereabouts-api")
 data class VideoBookingMigrateResponse(
```
matt-crowson-moj, line 70: Bringing inline with whereabouts-api types.

timharrison-moj (approved):
LGTM

PR #201, opened 2024-09-23
BAVL-355 wiring in telemetry client for capturing migration progress/success/failure.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/handlers/MigrateVideoBookingEventHandler.kt
```diff
@@ -30,12 +33,17 @@ class MigrateVideoBookingEventHandler(
         migrateVideoBookingService.migrate(bookingToMigrate)
       }
         .onFailure {
-          // TODO temporary telemetry details
-          telemetryService.capture("Failed migration of booking ${event.additionalInformation.videoBookingId} from whereabouts-api")
+          telemetryService.track(MigratedBookingFailureTelemetryEvent(event.additionalInformation.videoBookingId))
```
matt-crowson-moj, line 36: _Note: the exception is not being logged here ... should it be?_

timharrison-moj (approved):
LGTM. Should give us visibility - or the ability to look back for failures.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/handlers/MigrateVideoBookingEventHandler.kt
timharrison-moj: Is the onFailure() catching the exception? Just wondering if the message will go onto the DLQ?

matt-crowson-moj: It gets re-thrown so yes it will (still) go onto the DLQ, the `onFailure` is here so we can record it as telemetry.

PR #203, opened 2024-09-24
BAVL-357 increasing the max allowed comments length for migration of old BVLS records. The limit of 400 chars via endpoints remains in place.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #204, opened 2024-09-25
BAVL-342 the event details don't match up with that of the event raised by whereabouts.

matt-crowson-moj:
Easier to change our side instead of whereabouts ...


timharrison-moj (approved):
LGTM

PR #205, opened 2024-09-26
BAVL-342 handle unknown courts and probation in the comments and stop unrecoverable errors from going onto DLQ to stop unnecessary retries.

matt-crowson-moj:
The change includes the following:

- Prepends the free text court or probation team to the comments so we have a record of it during migration.
- Stops unrecoverable migration errors from going on the DLQ as retrying will not resolve it - _note we still get telemetry details recorded_.


timharrison-moj (approved):
LGTM

PR #207, opened 2024-09-27
BAVL-43 add config to clean refresh the preprod DB to avoid Flyway is sues whilst still testing.

matt-crowson-moj:


timharrison-moj (dismissed):
LGTM

PR #213, opened 2024-09-30
BAVL-43 handle OTHER court code and trim history comments to 1000 characters.

matt-crowson-moj:
A trial migration raised a couple of issues:

- `other` court codes
- `comments` too long for booking history

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/migration/MigrateVideoBookingService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/migration/MigrateVideoBookingService.kt
@@ -60,7 +60,7 @@ class MigrateVideoBookingService(
       )
     }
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/migration/MigrateVideoBookingService.kt
```diff
@@ -69,7 +69,7 @@ class MigrateVideoBookingService(
 
     val migratedBooking = VideoBooking.migratedCourtBooking(
       court = court,
-      comments = if (court.isUnknown()) "Court ${bookingToMigrate.courtName}\n".plus(bookingToMigrate.comment) else bookingToMigrate.comment,
+      comments = if (court.isUnknown()) "Court ${bookingToMigrate.courtName ?: "UNKNOWN"}\n".plus(bookingToMigrate.comment) else bookingToMigrate.comment,
```
matt-crowson-moj, line 72: Court name can be null with court is OTHER

PR #216, opened 2024-10-01
BAVL-43 change migration process to use prison-api instead of prisoner-search for prisoner lookup, prisoner-search does not return inactive bookings when looking up by booking id.

matt-crowson-moj:
We learnt when running migrations looking up inactive bookings by booking ID to get a prisoners number `prisoner-search-api` does not return anything.

To get the desired behavior (that is to find both active and inactive bookings by booking ID) we have had to change to use `prison-api` instead.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/migration/MigrationClient.kt
```diff
@@ -23,26 +21,30 @@ import java.time.LocalTime
...
 ) {
   companion object {
     private val log = LoggerFactory.getLogger(this::class.java)
   }
 
-  fun getPrisoner(bookingId: Long): Prisoner? =
-    prisonerSearchApiWebClient.post()
-      .uri("/prisoner-search/booking-ids")
-      .bodyValue(BookingIds(listOf(bookingId)))
+  // Unsure how effective this will be as requests will spread out over the pods unless we reduce number of pods.
```
matt-crowson-moj, line 33: Adding caching though not really sure how effective this will be due to events being handle over N pods.

PR #221, opened 2024-10-02
BAVL-364: Only allow external users to fetch a booking if the prison is self-service

smcveigh941:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/security/VideoBookingAccess.kt
```diff
@@ -11,11 +11,15 @@ import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service.UserType
  */
 fun checkVideoBookingAccess(externalUser: User, booking: VideoBooking) {
   if (externalUser.isUserType(UserType.EXTERNAL)) {
+    if (!booking.prisonIsEnabledForSelfService()) {
```
matt-crowson-moj, line 14: The only thing to be wary of here is appointments (at the moment) are lazily loaded on bookings so this must be called in the context of a `@Transactional` annotation.

Personally I'd be happy for the appointments to be eager on VideoBooking.

matt-crowson-moj (approved):
LGTM - one minor comment/suggestion ...

PR #222, opened 2024-10-03
BAVL-362 view for events used for CSV extraction should not restrict to just enabled courts and probation teams, we need to be able to view both enabled and disabled.

matt-crowson-moj:
This change fixes an issue noticed during migration testing.

The number of events (used for CSV extraction) in BVLS did not match up with the numbers in old BVL.  This is because we were filtering out disabled courts and probation team events in our view DDL.


timharrison-moj (approved):
LGTM

PR #224, opened 2024-10-04
BAVL-371 first part of 2 changes. The carries through free text court/probation team descriptions and stores on the migrated bookings for migrated records.

matt-crowson-moj:
The is the first of 2 PR's.

This change adds a new migration field to video bookings to capture free text court and probation team names for those cases where we do not have a matching court or probation team.

This new field needs to be included in the CSV extract when present (and not UNKNOWN which is what we are currently showing).  This will be dealt with in the next PR.

### DIFF
--- a/src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
+++ b/src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
@@ -190,7 +190,8 @@ CREATE TABLE video_booking
     created_time              timestamp    NOT NULL,
     amended_by                varchar(100),
     amended_time              timestamp,
-    migrated_video_booking_id bigint
+    migrated_video_booking_id bigint,
+    migrated_description      varchar(50)
 );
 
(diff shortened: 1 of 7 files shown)


src/main/resources/migrations/common/V2024.04.22__baseline_tables.sql
```diff
@@ -190,7 +190,8 @@ CREATE TABLE video_booking
     created_time              timestamp    NOT NULL,
     amended_by                varchar(100),
     amended_time              timestamp,
-    migrated_video_booking_id bigint
+    migrated_video_booking_id bigint,
+    migrated_description      varchar(50)
```
matt-crowson-moj, line 194: This should be big enough.  It is the same size as the old service [here](https://github.com/ministryofjustice/whereabouts-api/blob/main/src/main/resources/db/migration/V24__normalise_video_link_booking.sql#L8).

PR #226, opened 2024-10-07
BAVL-371 carry through the migrated description if present (to the UI) for migrated bookings.

matt-crowson-moj:
When a migrated booking has a free text description for the court or probation team this needs to be carried through to callers of the API e.g. front end.

Previously this would just have had a description of 'uknown'.


timharrison-moj (approved):
LGTM

PR #227, opened 2024-10-07
BAVL-372 external users can no longer see UNKNOWN courts/probation bookings created by prison users.

matt-crowson-moj:
After running a migration it became apparent by tweaking the URL when viewing a booking external users can see UNKNOWN bookings made by prison users.  This change now prevents that from happening.


timharrison-moj (approved):
LGTM

PR #235, opened 2024-10-14
BAVL-379 initial commit for enabling the new BVLS service in whereabouts-api

matt-crowson-moj:


migration/rollout_new_bvls_service.sh
```diff
@@ -0,0 +1,31 @@
...
+		break;;
+	[nN] ) echo exiting...;
+		exit;;
+	* ) echo invalid response;;
+esac
+
+done
+
+# Each service that needs changes applying should be added as a separate file and sourced below.
+source ./whereabouts-api.sh "$ENVIRONMENT"
```
matt-crowson-moj, line 29: Plan here is we can just add the other namespaces as separate files.  Just need to pass the `$ENVIRONMENT` through.

migration/whereabouts-api.sh
```diff
@@ -0,0 +1,31 @@
...
+  LISTEN_FOR_COURT_EVENTS: "false"
+  SEND_VIDEO_BOOKING_MIGRATE_EVENTS: "true"
+EOF
+
+# Apply the secret
+kubectl apply -f ./$WHEREABOUTS_SECRETS_FILE
+echo "Applied secrets to $WHEREABOUTS_NAMESPACE"
+
+kubectl -n "$WHEREABOUTS_NAMESPACE" get secrets feature-toggles -o json | jq -r ".data | map_values(@base64d)"
+
```
timharrison-moj, line 26: should be singular? `secrets`

timharrison-moj (approved):
LGTM

migration/whereabouts-api.sh
matt-crowson-moj: Both `secret` and `secrets` do the same 🤷‍♂️

PR #242, opened 2024-10-16
BAVL-383 removing the prisoner profile script, going down the PR route instead.

matt-crowson-moj:


migration/rollout-new-bvls-service.sh
```diff
@@ -34,9 +34,8 @@ esac
 done
 
 # Each service that needs changes applying should be added as a separate file and sourced below.
-#source whereabouts-api.sh "$ENVIRONMENT" out
```
matt-crowson-moj, line 37: Oops

PR #246, opened 2024-10-18
BAVL-238: Use court hearing/probation meeting description rather than code when propagating comments to appointment

smcveigh941:


matt-crowson-moj (approved):
LGTM

PR #249, opened 2024-10-19
BAVL-388: IP whitelist ranges

smcveigh941:


helm_deploy/hmpps-book-a-video-link-api/values.yaml
```diff
@@ -68,7 +68,6 @@ generic-service:
   allowlist:
     groups:
       - internal
-      - prisons
```
timharrison-moj, line 71: Should cloud-platform IPs be on this list, so the UI can call in?  Or does the group `internal` cover that?

timharrison-moj (approved):
LGTM

PR #251, opened 2024-10-21
BAVL-259 capturing metrics for amendment of court and probabtion bookings.

matt-crowson-moj:
Adds telemetry capture for court and probation booking amendments.


timharrison-moj (approved):
LGTM

PR #252, opened 2024-10-22
BAVL-260 capturing metrics for cancellation of court and probation bookings.

matt-crowson-moj:
This change introduces two new metrics for Court and probation cancellations.

There is also a bit of refactoring that has taken place as a result of introducing them.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/telemetry/ProbationBookingCreatedTelemetryEvent.kt
```diff
@@ -17,24 +15,8 @@ class ProbationBookingCreatedTelemetryEvent(private val booking: VideoBooking) :
...
-      "team_code" to booking.probationTeam!!.code,
-      "meeting_type" to booking.probationMeetingType!!,
-      "prison_code" to booking.prisonCode(),
-      "location_key" to booking.appointment().prisonLocKey,
-      "start" to booking.appointment().start().toIsoDateTime(),
-      "end" to booking.appointment().end().toIsoDateTime(),
-      "cvp_link" to (booking.videoUrl != null).toString(),
-    )
+  // Probation bookings are only created by probation users
+  override fun properties() = booking.commonProperties().plus("created_by" to "probation")
```
timharrison-moj, line 19: I like this booking.commonProperties().plus(xxx) thing ..

PR #255, opened 2024-10-24
BAVL-43 slight improvement to logging and adding temporary migration controller to assist with debugging.

matt-crowson-moj:
Improving telemetry failure logging around offender booking id not found.

Also added a temporary migration controller so we can migrate an individual booking on a need only basis e.g. something fails to migrate but we can easily retry it.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/migration/MigrateVideoBookingService.kt
```diff
@@ -43,7 +43,7 @@ class MigrateVideoBookingService(
 
   private fun migrateCourt(bookingToMigrate: VideoBookingMigrateResponse): VideoBooking {
     val prisonerNumber = mappingService.mapBookingIdToPrisonerNumber(bookingToMigrate.offenderBookingId)
-      ?: throw MigrationException("Unable to find prisoner number for booking ${bookingToMigrate.videoBookingId}")
+      ?: throw MigrationException("Unable to find prisoner number for offender booking ID ${bookingToMigrate.offenderBookingId} for court booking ${bookingToMigrate.videoBookingId}")
```
matt-crowson-moj, line 46: Saves having to look up the video booking, now has all we need to check i.e. the offender booking id.

PR #264, opened 2024-11-11
BAVL-405 increase the async timeout to one minute for CSV downloads.

matt-crowson-moj:
The CSV download was failing to download a years worth of data on migrated bookings in preprod.

Increasing the async timeout to see if this helps alleviate the issue.  Also added some progress logging for downloads.


timharrison-moj (approved):
LGTM
Might be a client-side timeout too?

PR #266, opened 2024-11-13
BAVL-390 temporarily disable processing of prisoner appointment cancelled event due to conflict with A&A causing the event to be raised when amend a booking.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #267, opened 2024-11-13
BAVL-390 re-enable processing of prisoner appointment cancelled event but now ignore if appointments serivce is rolled out for the prison raising the event.

matt-crowson-moj:
This change undoes this PR [here](https://github.com/ministryofjustice/hmpps-book-a-video-link-api/pull/266) and adds some extra logic/processing.

If the event to be processed is at a prison where A&A is rolled out the event will be completely ignored.  This avoids any conflicts with the event being raised by the A&A service.  If an appointment related to a booking is changed or cancelled in A&A then we will know anyway as it talks directly to our service.

Also caching calls to the A&A endpoint for rollout check as things don't change much here.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/handlers/PrisonerVideoAppointmentCancelledEventHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/handlers/PrisonerVideoAppointmentCancelledEventHandler.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service.events.handl
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Component
 import org.springframework.transaction.annotation.Transactional
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.client.activitiesappointments.ActivitiesAppointmentsClient
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.entity.HistoryType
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/handlers/PrisonerVideoAppointmentCancelledEventHandler.kt
```diff
@@ -46,6 +49,8 @@ class PrisonerVideoAppointmentCancelledEventHandler(
       )
 
       if (activeAppointments.size == 1) {
+        log.info("PRISONER VIDEO APPOINTMENT CANCELLED: processing $event")
```
matt-crowson-moj, line 52: Lets log it as well for now ...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/handlers/PrisonerVideoAppointmentCancelledEventHandler.kt
```diff
@@ -70,6 +75,8 @@ class PrisonerVideoAppointmentCancelledEventHandler(
     }
   }
 
+  private fun appointmentsNotManagedExternallyAt(prisonCode: String) = !activitiesAppointmentsClient.isAppointmentsRolledOutAt(prisonCode)
```
matt-crowson-moj, line 78: Not sure on the function name, meh!

timharrison-moj (approved):
Great stuff. LGTM.

PR #268, opened 2024-11-14
BAVL-390 minor change to the telemetry content for bookings cancelled by ServiceUser on the back of event processing, it is still a prison user who made the change.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #270, opened 2024-11-14
BAVL-413 alter approach checking for published messages so can run integration test in the CI pipeline.

matt-crowson-moj:
Instead of checking the actual queues for published messages we are waiting and checking the publisher is called instead.

This seems to be far more reliable/stable and runs every time on my local machine.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/resource/VideoLinkBookingIntegrationTest.kt
```diff
@@ -180,13 +180,21 @@ class VideoLinkBookingIntegrationTest : SqsIntegrationTestBase() {
     notifications.isPresent("g@g.com", NewCourtBookingPrisonCourtEmail::class, persistedBooking)
     notifications.isPresent(COURT_USER.email!!, NewCourtBookingUserEmail::class, persistedBooking)
 
-    2.messagesShouldBePublished {
+    await untilAsserted {
```
matt-crowson-moj, line 183: Far more stable.

timharrison-moj (approved):
LGTM

PR #271, opened 2024-11-15
BAVL-414 add missing health checks for dependent services.

matt-crowson-moj:
We were missing health checks for two dependent services, now added.

Also added some logging for the API clients.


timharrison-moj (approved):
LGTM

PR #283, opened 2024-11-22
BAVL-423 going back to previous version of the DPS gradle plugin as 6.1.0 breaks the Open API Swagger docs.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #287, opened 2024-11-25
Disable the cronjob at the weekend in dev, as deployments are down and we are getting alerts

smcveigh941:


timharrison-moj (approved):
LGTM

PR #289, opened 2024-11-29
BAVL-449 Until now prod in the migration scripts has been disabled to avoid potential mistake with running a migration in production.

matt-crowson-moj:


migration/rollback.sh
```diff
@@ -13,12 +13,6 @@ fi
 
 shopt -s nocasematch
 
-# Temporarily disable any prod runs
```
matt-crowson-moj, line 16: We can take this out now.

PR #295, opened 2024-12-02
BAVL-450 peek is an intermediate operation and should not be used for population, predominantly for debug.

matt-crowson-moj:
Replace original (buggy) cache approach with the genuine Spring cache.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/CsvDataExtractionService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/CsvDataExtractionService.kt
@@ -51,11 +51,8 @@ class CsvDataExtractionService(
   }
 
   private fun writeCourtBookingsToCsv(events: Stream<VideoBookingEvent>, csvOutputStream: OutputStream): Int {
-    val locationsByPrisonCode = mutableMapOf<String, List<Location>>()
-
     val courtEvents = events
-      .peek { locationsByPrisonCode.getOrPut(it.prisonCode) { locationsInsidePrisonClient.getNonResidentialAppointmentLocationsAtPrison(it.prisonCode) } }
-      .map { CourtBookingEvent(it, locationsByPrisonCode) }
+      .map { CourtBookingEvent(it, locationsInsidePrisonClient.getNonResidentialAppointmentLocationsAtPrison(it.prisonCode).toSet()) }
       .asSequence()
 
     var counter = 0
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/CsvDataExtractionService.kt
```diff
@@ -51,11 +51,8 @@ class CsvDataExtractionService(
   }
 
   private fun writeCourtBookingsToCsv(events: Stream<VideoBookingEvent>, csvOutputStream: OutputStream): Int {
-    val locationsByPrisonCode = mutableMapOf<String, List<Location>>()
-
     val courtEvents = events
-      .peek { locationsByPrisonCode.getOrPut(it.prisonCode) { locationsInsidePrisonClient.getNonResidentialAppointmentLocationsAtPrison(it.prisonCode) } }
-      .map { CourtBookingEvent(it, locationsByPrisonCode) }
+      .map { CourtBookingEvent(it, locationsInsidePrisonClient.getNonResidentialAppointmentLocationsAtPrison(it.prisonCode).toSet()) }
```
matt-crowson-moj, line 55: The locations endpoint is cached so it is safe to call here.  The previous code was done before we had any caching.

Also peek should not be used here as it is not a terminating operation where as map is on a Sequence.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/CsvDataExtractionService.kt
```diff
@@ -165,7 +159,7 @@ data class CourtBookingEvent(
   val preLocationName: String?,
   val postLocationName: String?,
 ) {
-  constructor(vbh: VideoBookingEvent, locations: Map<String, List<Location>>) : this(
+  constructor(vbh: VideoBookingEvent, locations: Set<Location>) : this(
```
matt-crowson-moj, line 162: Use of Set will guarantee unique locations supplied.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/resource/CsvDataExtractionIntegrationTest.kt
```diff
@@ -11,12 +11,13 @@ import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.helper.contains
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.helper.pentonvilleLocation
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.integration.IntegrationTestBase
 import java.time.LocalDate
+import java.util.UUID
 
 class CsvDataExtractionIntegrationTest : IntegrationTestBase() {
 
   @BeforeEach
   fun before() {
-    locationsInsidePrisonApi().stubNonResidentialAppointmentLocationsAtPrison(setOf(pentonvilleLocation.key), prisonCode = PENTONVILLE)
```
matt-crowson-moj, line 19: This was still set up on the premise we where using business keys and UUIDs.

PR #296, opened 2024-12-02
BAVL-463: Fix some data where Gartree bookings are not synced with nomis

smcveigh941:
Before we synced data from Whereabouts, the A&A team did a data fix where they accidentally cancelled some appointments at Gartree, and they reverted that delete [here](https://github.com/ministryofjustice/hmpps-activities-management-api/blob/main/src/main/resources/migrations/prod/V2024.10.31__gartree_appts_undelete.sql)

The trouble is, they did not also fix the data in Whereabouts, so we have migrated across the bad data. We are fixing the bad data here post-migration


timharrison-moj (approved):
LGTM

PR #305, opened 2024-12-05
BAVL-471 captilising location descriptions (were all uppercase from locations API) in locations returned via API and in emails.

matt-crowson-moj:
Whilst location formatting could be done in the UI, there are 3 codebases using the locations endpoint so this seems like the logic place to do the formatting.

Also emails will now also have formatted locations instead of all just uppercase for consistency.


timharrison-moj (approved):
LGTM

PR #307, opened 2024-12-05
BAVL-439 removal of all mirgation specfic code now that the migration of old bookings to new has been done in production.

matt-crowson-moj:
Goodbye migration code.

Now that we have migrated the old BVLS service data into the re-platformed service we no longer need the migration specific code.


utility/display-maintenance-screen.sh
matt-crowson-moj, line 1: @timharrison-moj FYI

utility/remove-maintenance-screen.sh
matt-crowson-moj, line 1: @timharrison-moj FYI

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/nomismapping/NomisMappingClient.kt
```diff
@@ -13,16 +13,6 @@ class NomisMappingClient(private val nomisMappingApiWebClient: WebClient) {
     private val log = LoggerFactory.getLogger(this::class.java)
   }
 
-  @Deprecated(message = "Can be removed when migration is completed")
```
matt-crowson-moj, line 16: @timharrison-moj we do not need this, this is for migration only.

timharrison-moj (approved):
LGTM

PR #308, opened 2024-12-06
BAVL-522 preparation PR for the clashing appointments fix. Pulling out common interface in availability check to combine data from multiple sources.

matt-crowson-moj:
Behavior should be the same as a result of this PR i.e. we are still only looking locally for clashing appointments.

Following PR will combine with appointment data not mastered in BVLS.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoAppointment.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoAppointment.kt
@@ -35,11 +35,19 @@ data class VideoAppointment(
 
   val appointmentType: String,
 
-  val prisonLocationId: UUID,
+  override val prisonLocationId: UUID,
 
   val appointmentDate: LocalDate,
 
-  val startTime: LocalTime,
+  override val startTime: LocalTime,
 
-  val endTime: LocalTime,
-)
+  override val endTime: LocalTime,
+) : AppointmentSlot
+
+interface AppointmentSlot {
+  val prisonLocationId: UUID
+
+  val startTime: LocalTime
+
+  val endTime: LocalTime
+}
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/resource/JobTriggerIntegrationTest.kt
```diff
@@ -152,14 +152,17 @@ class JobTriggerIntegrationTest : IntegrationTestBase() {
   }
 
   @Test
-  @Sql("classpath:integration-test-data/seed-migrated-bookings.sql")
+  @Sql("classpath:test_data/clean-all-data.sql", "classpath:integration-test-data/seed-migrated-bookings.sql")
   fun `should not send an email to the court to remind them to add the court hearing link if the booking has been migrated`() {
-    notificationRepository.findAll() hasSize 0
+    assumingThat(today().dayOfWeek in listOf(MONDAY, TUESDAY, WEDNESDAY, THURSDAY, SUNDAY)) {
+      notificationRepository.findAll() hasSize 0
```
matt-crowson-moj, line 158: Data was hanging around causing the first assert to fail.

Not 100% sure why added the clean specific to the endpoint to work around this for now.  Will come back to it.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoAppointment.kt
```diff
@@ -35,11 +35,19 @@ data class VideoAppointment(
...
 
-  val startTime: LocalTime,
+  override val startTime: LocalTime,
 
-  val endTime: LocalTime,
-)
+  override val endTime: LocalTime,
+) : AppointmentSlot
+
+interface AppointmentSlot {
```
matt-crowson-moj, line 47: Common interface.  I might move this in following PRs but this will work as a first cut.

timharrison-moj (commented):
LGTM (retrospectively!)

PR #312, opened 2024-12-09
BAVL-522 add missing brackets off ANDED date range queries for the CSV extract.

matt-crowson-moj:


timharrison-moj (approved):
LGTM. I think we should also try this in PREPROD - verify we only get events within the dates? I know you will anyway... don't know why I've started stating the obvious.

PR #313, opened 2024-12-09
BAVL-522 using timestamp instead of date_booked (date_booked is the same for all events related to a booking).

matt-crowson-moj:
Use timestamp (AKA creation timestamp) instead of date_booked for query.

date_booked would be the same for all events for a booking even if they occurred on different days.


timharrison-moj (approved):
LGTM

PR #314, opened 2024-12-10
BAVL-522 using timestamp instead of date_booked (date_booked is the same for all events related to a booking).

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #317, opened 2024-12-11
BAVL-531 filter out return to unit locations from calls to locations API as these are not bookable locations.

matt-crowson-moj:
This change resolves an issue whereby a user was able to make a booking into a location that is not a really bookable, `return to unit`.

As a result when an attempt was made to sync a booking with this location into NOMIS via A&A an error was being thrown preventing the appointment ever being synced.

I think strictly speaking this should not be returned by the locations API as a bookable location.  However for safety we are also filtering it out in our service.


timharrison-moj (approved):
lgtm

PR #320, opened 2024-12-16
Fix cache eviction - all entries should be evicted

smcveigh941:
`@CacheEvict` accepts several parameters:

`value`: Specifies the cache name(s) to operate on.
`allEntries` (default false): If true, all entries in the cache are evicted.
`key`: Specifies the cache key(s) to evict, if you want to evict specific entries.

With `allEntries = false` (default), the cache won't be evicted because no key is provided and Spring doesn't know which entry to evict.


timharrison-moj (approved):
LGTM

PR #326, opened 2024-12-20
Appointment cancelled event for an appointment which has already happened is a No-Op

smcveigh941:


timharrison-moj (approved):
LGTM

timharrison-moj (approved):
LGTM

PR #330, opened 2025-01-14
BAVL-557 use new attribute on A&A prison rollout plan to determine if a prison is live or not.

matt-crowson-moj:
There is now a dedicated attribute on the prison rollout plan to determine if a prison is live or not.

This is now the preferred way of working this out.


timharrison-moj (approved):
LGTM. 
Timing though - we should probably wait for the A&A change to get into production?

matt-crowson-moj:
> LGTM. Timing though - we should probably wait for the A&A change to get into production?

@timharrison-moj it is already in production

PR #331, opened 2025-01-15
BAVL-525 additional columns for CSV, hearing type for court, meeting type for probation and the user.

matt-crowson-moj:
Additional CSV extract columns:

For court bookings - hearing type and user responsible for the event (create/update/delete).

For probation bookings - meeting type and user responsible for the event (create/update/delete).


timharrison-moj (approved):
LGTM

PR #334, opened 2025-01-16
BAVL-570 added in some simple validation to ensure booking is in correct state before recording any history.

matt-crowson-moj:
Added some basic validation checks prior to creating any booking history for a video booking.


timharrison-moj (approved):
LGTM.

PR #335, opened 2025-01-17
BAVL-571 do not automatically add extra information to a bookings comments, only users should be adding/changing comments.

matt-crowson-moj:
The original intention of prefixing comments was so prison users could see information about court/probation meetings in other areas where this information would not be visible e.g. NOMIS, 7 day schedule etc.

This does however cause confusion as this is not something that has been entered by the user that created or amended a bookings comments.


timharrison-moj (approved):
LGTM

PR #342, opened 2025-01-28
BAVL-840: Add createdBy and amendedBy fields to the ScheduleItem response model

smcveigh941:
Needed by the cancellation screen which displays cancelledBy column (which is amendedBy)


matt-crowson-moj (approved):
LGTM

PR #347, opened 2025-01-31
BAVL-602 refactoring booking type identifcation to use BookType.

matt-crowson-moj:
The way we were identifying a booking type was getting a bit messy.

Hopefully this change makes a bit easier and will make introducing introducing other types going forwards e.g. OTHER.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
```diff
@@ -190,77 +189,6 @@ class VideoBooking private constructor(
         createdBy = createdBy,
         createdByPrison = createdByPrison,
       )
-
-    fun migratedCourtBooking(
```
matt-crowson-moj, line 194: I missed then when ripping out migration code.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookingHistoryService.kt
```diff
@@ -32,10 +34,10 @@ class BookingHistoryService(private val bookingHistoryRepository: BookingHistory
     BookingHistory(
       videoBookingId = booking.videoBookingId,
       historyType = historyType,
-      courtId = booking.court?.courtId.takeIf { booking.isCourtBooking() },
-      hearingType = booking.hearingType.takeIf { booking.isCourtBooking() },
-      probationTeamId = booking.probationTeam?.probationTeamId.takeUnless { booking.isCourtBooking() },
```
matt-crowson-moj, line 37: Lines 37 and 38 would not have worked if we introduced another type ...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookingFacade.kt
```diff
@@ -54,7 +56,7 @@ class BookingFacade(
 
   fun create(bookingRequest: CreateVideoBookingRequest, createdBy: User): Long {
     require(createdBy is PrisonUser || availabilityService.isAvailable(bookingRequest)) {
-      if (bookingRequest.bookingType == BookingType.COURT) {
+      if (bookingRequest.bookingType == RequestBookingType.COURT) {
         "Unable to create court booking, booking overlaps with an existing appointment."
```
timharrison-moj, line 60: Wondering why this is the only place which uses an alias of BookingType?

timharrison-moj (approved):
LGTM.
One question raised.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookingFacade.kt
matt-crowson-moj: Because this is was already using a BookingType object but taken from the model package.  We have one in Domain.Perhaps ModelBookingType would be better?

PR #356, opened 2025-02-14
BAVL-617 additional details are not mandatory for probation bookings.

matt-crowson-moj:
Additional details are no longer mandatory or probation bookings.

When they are provided name and email must be provided.  Contact number is optional.


timharrison-moj (approved):
LGTM

PR #358, opened 2025-02-14
BAVL-625 resolve deprecation warnings in tests.

matt-crowson-moj:


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/resource/VideoLinkBookingControllerTest.kt
```diff
@@ -69,29 +71,46 @@ import java.util.Optional
...
+  }
+
+  @MockitoBean
   private lateinit var bookingFacade: BookingFacade
 
-  @MockBean
+  @MockitoBean
   private lateinit var requestBookingService: RequestBookingService
 
-  @SpyBean
```
matt-crowson-moj, line 82: So here you are supposed to use MockitoSpyBean.  However there is a known issue with this annotation so I have added a test config to allow me to override the service directly instead.

timharrison-moj (approved):
LGTM

PR #363, opened 2025-02-20
BAVL-628 change to factor in existing booking. If one is supplied then results will still show the existing bookings time as available in the results.

matt-crowson-moj:
This change allows for the amend journey where we still want to see the times in the allowable slots for the existing journeys appointments prior to changing it.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookedLocationsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookedLocationsService.kt
@@ -1,12 +1,21 @@
 package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service
 
+import jakarta.persistence.EntityNotFoundException
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookedLocationsService.kt
```diff
@@ -16,40 +25,87 @@ class BookedLocationsService(
...
+
 data class BookedLocation(val location: Location, val startTime: LocalTime, val endTime: LocalTime)
 
 class BookedLocations(booked: Collection<BookedLocation>) {
 
-  private val locations = booked.groupBy { it.location.key }
+  private val locations = booked.groupBy { it.location.dpsLocationId }
 
-  fun isBooked(location: Location, startTime: LocalTime, endTime: LocalTime) = locations[location.key]?.any { isTimesOverlap(it.startTime, it.endTime, startTime, endTime) } ?: false
+  fun isBooked(location: Location, startTime: LocalTime, endTime: LocalTime) = locations[location.dpsLocationId]?.any { isTimesOverlap(it.startTime, it.endTime, startTime, endTime) } ?: false
```
matt-crowson-moj, line 110: Changed to use the ID instead of key for the location, although using the key still works.

Should be avoiding use of the key really where possible.

PR #366, opened 2025-02-21
BAVL-635 include additional details on GET video booking if there are any for a booking.

matt-crowson-moj:
This is support the new probation journey where we can capture optional additional details.

We need to be able to pull them back as well e.g. for amend journeys.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/resource/VideoLinkBookingControllerTest.kt
```diff
@@ -89,12 +90,16 @@ class VideoLinkBookingControllerTest {
...
+
     @Bean
     fun videoLinkBookingsService() = VideoLinkBookingsService(
       videoBookingRepository(),
       referenceCodeRepository(),
       videoAppointmentRepository(),
       locationsInsidePrisonClient(),
+      additionalBookingDetailRepository(),
     )
   }
```
timharrison-moj, line 104: What's the reason for these being @Bean declarations? In tests usually enough to declare repos as mocks and construct the service in the test class? Just interested if there's any additional benefit in Beans.

timharrison-moj (approved):
LGTM
I think this is a non-breaking change, as optional on response object

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/resource/VideoLinkBookingControllerTest.kt
matt-crowson-moj: See here - https://github.com/ministryofjustice/hmpps-book-a-video-link-api/pull/358

When I made the original change to get rid of the deprecation warnings there was a knock on effect due to a known bug with the replacement Mockito annotations i.e. this is a workaround.

PR #367, opened 2025-02-21
BAVL-628 will start looking for available locations from now going forwards for requests for today.

matt-crowson-moj:
The change will adjust the point in time we start looking for available locations if the request is for today an we are beyond the regime start time.

E.g. 

With regime start time = 8:00AM and date = today

Adjusted start times below:

10:00AM -> 10:15
10:01AM -> 10:30
10:15AM -> 10:30
10:16AM -> 10:45

and so on ...

We may want to soften these a little e.g. 10:01 becoming 10:30 might a bit extreme.


timharrison-moj (approved):
LGTM

PR #369, opened 2025-02-24
BAVL-628 including checks for those rooms that are decorated.  The logic in the LocationAttribute entity now needs extending.

matt-crowson-moj:
So the fundamental flow is in place with this PR.

There may be further work to do with regards to ordering/prioritisation of rooms.

More attention/work is still needed to the rules in the Domain for the LocationAttribute entity e.g. scheduled.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/AvailableLocationsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/AvailableLocationsService.kt
@@ -1,5 +1,6 @@
 package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service
 
+import jakarta.persistence.EntityNotFoundException
 import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
 import org.springframework.transaction.annotation.Transactional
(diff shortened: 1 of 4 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/AvailableLocationsService.kt
```diff
@@ -107,15 +111,88 @@ class AvailableLocationsService(
...
+        )
       }
     }
 
     return true
   }
 }
+
+@Service
+@Transactional(readOnly = true)
```
matt-crowson-moj, line 139: The intent on a later PR is to move all of the location related services into a sub-package e.g. locations.  As there are now a few locations related services.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/AvailableLocationsService.kt
```diff
@@ -51,19 +58,16 @@ class AvailableLocationsService(
...
             request.fallsWithinSlotTime(meetingStartTime) &&
-            location.allowsByAnyRuleOrSchedule(request)
+            location.allowsByAnyRuleOrSchedule(request, meetingStartTime)
           ) {
-            // TODO need to check against the room decoration/schedules
             add(
               AvailableLocation(
-                // TODO is it possible for the description to be null?
-                name = location.description ?: "UNKNOWN",
+                name = location.description ?: location.key,
```
matt-crowson-moj, line 65: It feels like putting the key here should suffice on the off chance there is no description for a location.

timharrison-moj (approved):
LGTM

PR #373, opened 2025-02-25
BAVL-628 applying restrictions to dedicated booking types, dedicated booking codes/teams and shared roomes. Does not include schedules.

matt-crowson-moj:
These changes reduce the room choices based on the room decorations bar room schedules.

Schedules are still yet to be done.


timharrison-moj (approved):
LGTM.

PR #374, opened 2025-02-26
Probation changes feature toggle as a secret

smcveigh941:
Having this as a secret which is shared between the UI and the API means that the toggle can be turned on in both places at the same time


helm_deploy/hmpps-book-a-video-link-api/values.yaml
```diff
@@ -64,6 +63,8 @@ generic-service:
       HMPPS_SQS_QUEUES_BVLS_QUEUE_NAME: "sqs_queue_name"
     sqs-domain-event-dlq-secret:
       HMPPS_SQS_QUEUES_BVLS_DLQ_NAME: "sqs_queue_name"
+    feature-toggles:
+      FEATURE_MASTER_VLPM_TYPES: "FEATURE_MASTER_VLPM_TYPES"
```
timharrison-moj, line 68: Have you already created in all envs?

timharrison-moj (approved):
LGTM

PR #376, opened 2025-02-26
BAVL-628 adding logic for probabation availability with schedules.

matt-crowson-moj:
Changes to support room/location availability for probation location/rooms decorated with schedules.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
@@ -13,6 +13,7 @@ import jakarta.persistence.OneToMany
 import jakarta.persistence.OneToOne
 import jakarta.persistence.Table
 import org.hibernate.Hibernate
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.common.between
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.common.isBetween
 import java.time.LocalDate
 import java.time.LocalDateTime
@@ -97,7 +98,7 @@ class LocationAttribute(
       LocationUsage.SHARED -> AvailabilityStatus.SHARED
       LocationUsage.PROBATION -> {
         if (allowedParties == null) {
-          AvailabilityStatus.PROBATION
+          AvailabilityStatus.PROBATION_ANY
         } else {
           if (isPartyAllowed(probationTeam.code)) {
             AvailabilityStatus.PROBATION_TEAM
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
```diff
@@ -111,20 +112,26 @@ class LocationAttribute(
...
+              probationTeam,
+              startingOnDateTime,
+            )
+          }
+        ) {
+          return AvailabilityStatus.PROBATION_TEAM
         }
+        if (any { schedule -> schedule.isAvailableForAnyProbationTeam(startingOnDateTime) }) return AvailabilityStatus.PROBATION_ANY
+        return AvailabilityStatus.NONE
       }
```
timharrison-moj, line 133: I like this, reads well.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
```diff
@@ -149,9 +156,7 @@ class LocationAttribute(
     }
   }
 
-  private fun Int.between(from: Int, to: Int) = this in from..to
-
-  private fun isPartyAllowed(party: String) = allowedParties.isNullOrBlank() || allowedParties.replace(" ", "").split(",").contains(party)
+  private fun isPartyAllowed(party: String) = allowedParties.orEmpty().replace(" ", "").split(",").contains(party)
 }
```
timharrison-moj, line 160: `party` being either a `courtCode` or `probationTeamCode`

matt-crowson-moj: Yes.  Can try and rename if not clear enough.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationScheduleTest.kt
```diff
@@ -0,0 +1,107 @@
...
+    @Test
+    fun `should be available for probation team on Monday at noon`() {
+      val schedule = schedule(locationUsage = LocationUsage.PROBATION, allowedParties = "PROBATION_TEAM")
+
+      schedule.isAvailableForProbationTeam(probationTeam(code = "PROBATION_TEAM"), mondayNoon) isBool true
+    }
+
+    @Test
+    fun `should be available for probation team on Monday out of hours`() {
+      val schedule = schedule(locationUsage = LocationUsage.PROBATION, allowedParties = "PROBATION_TEAM")
```
timharrison-moj, line 32: This is only testing that the schedule reports SHARED (available) for midnight, but not that the prison opening hours of 8am - 18pm will support a meeting, as that is the higher-level availability check?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationScheduleTest.kt
```diff
@@ -0,0 +1,107 @@
...
+
+      schedule.isAvailableForAnyProbationTeam(mondayMidnight) isBool true
+    }
+
+    @Test
+    fun `should not be available for any probation team on Monday at noon`() {
+      val schedule = schedule(locationUsage = LocationUsage.PROBATION, allowedParties = "PROBATION_TEAM")
+
+      schedule.isAvailableForAnyProbationTeam(mondayNoon) isBool false
+    }
```
timharrison-moj, line 63: Good set of tests. I'm sure there will be other variations related to multiple schedules in place, any one of which might allow? But probably a higher level test.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationScheduleTest.kt
matt-crowson-moj: Yes. If the config/prison setup was part of the Prison entity then we could probably handle this here at a lower level.  I don't want to have to pass it down from the service .

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationScheduleTest.kt
matt-crowson-moj: This is s tricky one.  The number of permutations is high, so the balance is a fine line.

PR #378, opened 2025-02-27
BAVL-658 deriving the PCVL link using room attributes for probation bookings.

matt-crowson-moj:
This covers off most what I think is needed with regards to showing the video URL for probation bookings.

This is not behind any feature toggle, if a room is decorated then it will show the link in the existing service, I don't think this really matters to be honest.

@timharrison-moj I am unsure on this point on the Jira ticket `The probation appointment will have a room` so I have not done this yet.


timharrison-moj (approved):
LGTM

PR #379, opened 2025-02-28
BAVL-658 adding missing transactional annotation, now need as reading location attributes which are lazy loaded.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #381, opened 2025-02-28
BAVL-658 resolve an issue whereby shared rooms were being picked in favour of dedicated rooms.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
```diff
@@ -107,32 +105,30 @@ class LocationAttribute(
...
+   * We need to look at the schedule as a whole to determine availability. A schedule on its own is not enough.
+   */
+  private fun getScheduleAvailability(probationTeam: ProbationTeam, dateAndTime: LocalDateTime): AvailabilityStatus {
+    val schedules = locationSchedule.filter { it.fallsOn(dateAndTime) }
+
+    // If there aren't any schedules which fall on the requested date and time then the default fallback is SHARED
+    if (schedules.isEmpty()) {
+      return AvailabilityStatus.SHARED
+    } else {
+      // If there are matches then look to see if any match the following
```
matt-crowson-moj, line 123: We know at this point we only have schedules that fall within the time window of interest given the above ergo we only need to check if the following apply here.

timharrison-moj (approved):
LGTM

PR #385, opened 2025-03-04
BAVL-664 intial commit to support creation of a decorated room.

matt-crowson-moj:


timharrison-moj (approved):
Very quick and thorough. Thanks. LGTM.

PR #398, opened 2025-03-07
BAVL-676 finishing location attribute and schedule checks for courts, this is still needed regardless of how rooms at searched.

matt-crowson-moj:
Initially I left this as a **TODO**.  Now the probation part is done I have filled that gap for courts at the Domain level.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
```diff
@@ -182,29 +180,42 @@ class LocationAttribute private constructor(
...
+
+      LocationUsage.SCHEDULE -> getScheduleAvailability(court, startingOnDateTime)
       else -> return AvailabilityStatus.NONE
     }
   }
 
+  /**
+   * We need to look at the schedules as a whole to determine availability. A schedule on its own is not enough.
+   */
+  private fun getScheduleAvailability(court: Court, dateAndTime: LocalDateTime): AvailabilityStatus {
```
matt-crowson-moj, line 204: This was the TODO part.  The logic is the same (but for courts) regardless of single room or multiple room model.

timharrison-moj (approved):
LGTM
Set's us up nicely for next week's brainteaser.

PR #399, opened 2025-03-10
BAVL-676 using bulk nomis mapping location lookup endpoint now that it is available in favour of multiple single calls.

matt-crowson-moj:
This replaces multiple single GET requests to the NOMIS mapping service with a single POST call now that a bulk lookup endpoint is available on the NOMIS mapping service.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/nomismapping/NomisMappingClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/nomismapping/NomisMappingClient.kt
@@ -1,11 +1,16 @@
 package uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.client.nomismapping
 
 import org.slf4j.LoggerFactory
+import org.springframework.cache.annotation.Cacheable
+import org.springframework.core.ParameterizedTypeReference
 import org.springframework.stereotype.Component
 import org.springframework.web.reactive.function.client.WebClient
 import org.springframework.web.reactive.function.client.WebClientResponseException
 import reactor.core.publisher.Mono
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.config.CacheConfiguration
 import java.util.UUID
+
+inline fun <reified T> typeReference() = object : ParameterizedTypeReference<T>() {}
 
 @Component
 class NomisMappingClient(private val nomisMappingApiWebClient: WebClient) {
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/nomismapping/NomisMappingClient.kt
```diff
@@ -25,6 +39,20 @@ class NomisMappingClient(private val nomisMappingApiWebClient: WebClient) {
     .block()
 }
 
+class DpsLocationsIds(locationIds: Collection<UUID>) {
```
matt-crowson-moj, line 42: Wrapping the IDs in a class means we can better support caching to ensure the order of the IDs are always in the same order.  If the same IDs were used but in a different order then it would be seen as an entirely new request.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/nomismapping/NomisMappingClient.kt
```diff
@@ -1,20 +1,34 @@
...
+
 @Component
 class NomisMappingClient(private val nomisMappingApiWebClient: WebClient) {
   companion object {
     private val log = LoggerFactory.getLogger(this::class.java)
   }
 
-  fun getNomisLocationMappingsBy(dpsLocationIds: Collection<UUID>): Collection<NomisDpsLocationMapping> = dpsLocationIds.mapNotNull { getNomisLocationMappingBy(it) }
+  @Cacheable(CacheConfiguration.NOMIS_MAPPING_CACHE_NAME)
+  fun getNomisLocationMappingsBy(dpsLocationIds: DpsLocationsIds) = nomisMappingApiWebClient
```
matt-crowson-moj, line 22: Decided this is a good caching candidate.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/config/CacheConfiguration.kt
```diff
@@ -22,30 +22,40 @@ class CacheConfiguration {
...
   @CacheEvict(value = [ROLLED_OUT_PRISONS_CACHE_NAME], allEntries = true)
-  @Scheduled(fixedDelay = 5, timeUnit = TimeUnit.MINUTES)
+  @Scheduled(fixedDelay = FIVE, timeUnit = TimeUnit.MINUTES)
   fun cacheEvictRolledOutPrisons() {
-    log.info("Evicting cache: $ROLLED_OUT_PRISONS_CACHE_NAME after 5 mins")
+    log.info("Evicting cache: $ROLLED_OUT_PRISONS_CACHE_NAME after $FIVE mins")
+  }
+
+  @CacheEvict(value = [NOMIS_MAPPING_CACHE_NAME], allEntries = true)
+  @Scheduled(fixedDelay = TWO, timeUnit = TimeUnit.HOURS)
```
matt-crowson-moj, line 57: Safe to cache this one in hours (I think).

PR #419, opened 2025-03-19
BAVL-696 fixing bug with prison users in new probabtion journey, not setting created by prison correctly.

matt-crowson-moj:
The CSV would have been wrong as a result of this bug.  Note: this does not affect the old journeys, only the new.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
@@ -14,6 +14,7 @@ import jakarta.persistence.OneToOne
 import jakarta.persistence.Table
 import org.hibernate.Hibernate
 import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.common.requireNot
+import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service.PrisonUser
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/CreateProbationBookingService.kt
```diff
@@ -51,8 +51,7 @@ class CreateProbationBookingService(
       probationTeam = probationTeam,
       probationMeetingType = request.probationMeetingType!!.name,
       comments = request.comments,
-      createdBy = createdBy.username,
-      createdByPrison = false,
```
matt-crowson-moj, line 55: Noticed this bug during testing.  In the new world prison users can create probation bookings.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
```diff
@@ -178,8 +178,8 @@ class VideoBooking private constructor(
       probationMeetingType = probationMeetingType,
       comments = comments,
       videoUrl = null,
-      createdBy = createdBy,
-      createdByPrison = createdByPrison,
+      createdBy = createdBy.username,
+      createdByPrison = createdBy is PrisonUser,
```
matt-crowson-moj, line 182: Pushed this down into the entity.  Will do the same for court bookings in a separate PR.

PR #424, opened 2025-03-24
BAVL-663 you can now add a schedule row to a location attribute even if the location attribute is not scheduled. It will have no effect.

matt-crowson-moj:
Relax adding a schedule row to a location attribute.  Can add even if the location attribute is not scheduled.  The schedule row(s) will have no effect on room selection.


timharrison-moj (approved):
LGTM. Thanks.

PR #426, opened 2025-03-25
BAVL-663 removed scheduled row constraint from domain but missed in service layer.

matt-crowson-moj:


timharrison-moj (approved):
LGTM. That'll do it!

PR #428, opened 2025-03-25
BAVL-718 changes to include new probation values in the new probation booking email (isolated to dev only).

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/BookingFacade.kt
```diff
@@ -44,10 +45,11 @@ class BookingFacade(
   private val emailService: EmailService,
   private val notificationRepository: NotificationRepository,
   private val outboundEventsService: OutboundEventsService,
-  private val locationsInsidePrisonClient: LocationsInsidePrisonClient,
+  private val locationsService: LocationsService,
```
matt-crowson-moj, line 48: Now using location service here as this provides what we need for extra room attributes to pass onto email.s

src/main/resources/application-dev.yml
```diff
@@ -1,3 +1,9 @@
 spring:
   flyway:
     locations: classpath:/migrations/common,classpath:/migrations/dev
+
+notify:
```
matt-crowson-moj, line 5: This is the new template with the additional place holder fields included in DEV only.

In all other environments it will carry on using the old email which does not expect the new placeholder values.

timharrison-moj: Would you be able to add these instructions in the rollout confluence page please?

timharrison-moj (approved):
LGTM

PR #432, opened 2025-03-26
BAVL-702: Remove the location from the booking request journey

smcveigh941:
- Removes the location from the request a booking API endpoint, and resulting emails
- This is not behind a feature toggle, so the locations will be removed from the resulting emails when this is deployed to prod, however nothing should break.
- The location fields from the UI will be removed from the UI when we toggle ON the new court and probation journeys


matt-crowson-moj (approved):
LGTM

PR #434, opened 2025-03-26
BAVL-689: Enable HMP Brixton

timharrison-moj:
This will allow room decoration to be applied, but the UI will disallow booking until we remove BXI from the FEATURE_GREY_RELEASE_PRISONS env var (secrets in K8s)


matt-crowson-moj (approved):
LGTM

PR #438, opened 2025-03-27
BAVL-732 suggested meeting end times should not go beyond the end of the prison working day 18:00.

matt-crowson-moj:
Meeting end times cannot go beyond the end of the prison working day.

When prompting suggested available times the end time must be before or at the end of the working day.


timharrison-moj (approved):
LGTM

PR #441, opened 2025-03-28
BAVL-717 including pre and post PVL's in amend court booking emails.

matt-crowson-moj:
This covers off all of the amend emails for court booking amendments.


timharrison-moj (approved):
LGTM

PR #451, opened 2025-04-04
BAVL-713 adding additional details and PVL to the probation cancellation emails.

matt-crowson-moj:
Adds the probation officers details and PVL to the probation cancellation emails (in DEV only).

### DIFF
--- a/src/main/resources/application-dev.yml
+++ b/src/main/resources/application-dev.yml
@@ -1,3 +1,13 @@
 spring:
   flyway:
     locations: classpath:/migrations/common,classpath:/migrations/dev
+
+notify:
+  templates:
+    probation:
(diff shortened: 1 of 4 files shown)


src/main/resources/application-dev.yml
```diff
@@ -1,3 +1,13 @@
 spring:
   flyway:
     locations: classpath:/migrations/common,classpath:/migrations/dev
+
+notify:
+  templates:
+    probation:
+      cancelled-booking:
```
matt-crowson-moj, line 8: Once this is tested in dev and the API has been pushed out we can switch over to the new templates in production.

timharrison-moj (approved):
LGTM

PR #461, opened 2025-04-10
BAVL-749: Patch appointments in A&A instead of deleting and re-creating them

smcveigh941:


matt-crowson-moj (approved):
LGTM - good stuff!

PR #462, opened 2025-04-10
BAVL-752 changed to use prison-api to determine last movement due to race condition with prisoner search indexing.

matt-crowson-moj:
Trying prison API to determine reason for last movement, the prisoner search index is not in sync with the event raised by NOMIS.


timharrison-moj (approved):
LGTM

PR #467, opened 2025-04-17
BAVL-788 fix schedule amend row when save row with no changes.

matt-crowson-moj:
I have manually tested this locally.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationSchedule.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationSchedule.kt
@@ -97,8 +97,7 @@ class LocationSchedule private constructor(
     require(startTime.isBefore(endTime)) { "The end time must come after the start time." }
 
     if (
-      locationAttribute.schedule().any {
-        it.locationScheduleId != locationScheduleId
+      locationAttribute.schedule().filterNot { it.locationScheduleId == this.locationScheduleId }.any {
         it.locationUsage == locationUsage &&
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationSchedule.kt
```diff
@@ -97,8 +97,7 @@ class LocationSchedule private constructor(
     require(startTime.isBefore(endTime)) { "The end time must come after the start time." }
 
     if (
-      locationAttribute.schedule().any {
-        it.locationScheduleId != locationScheduleId
```
matt-crowson-moj, line 101: Was missing ampersand of the the check.  This should have been using a filter {} anyway, so corrected.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationScheduleTest.kt
```diff
@@ -101,44 +101,6 @@ class LocationScheduleTest {
     }.message isEqualTo "The end time must come after the start time."
   }
 
-  @Test
-  fun `should fail to amend if causes a duplicate row`() {
```
matt-crowson-moj, line 105: This test was floored anyway (regardless of the bug).

Unfortunately we cannot unit test as it relies on ID checking.  We cannot set or change the IDs to be able to test this.

matt-crowson-moj: We lack integration tests in this area of the service.

PR #472, opened 2025-04-23
BAVL-778: Ignore cancelled appointments from A&A which were blocking rooms.

timharrison-moj:
This PR:
* Ignores cancelled appointments returned from A&A appointments search when checking whether a room is available.
* Wrote a failing test first - which failed - then implemented the change to make the test pass.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/locations/availability/BookedLocationsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/locations/availability/BookedLocationsService.kt
@@ -55,6 +55,8 @@ class BookedLocationsService(
 
     return when (activitiesAppointmentsClient.isAppointmentsRolledOutAt(prisonCode)) {
       true -> activitiesAppointmentsClient.getScheduledAppointments(prisonCode, date, nomisToDpsLocations.keys)
+        // Do not consider cancelled appointments
+        .filterNot { it.isCancelled }
         .mapNotNull { result ->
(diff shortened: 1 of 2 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/locations/availability/BookedLocationsService.kt
```diff
@@ -55,6 +55,8 @@ class BookedLocationsService(
 
     return when (activitiesAppointmentsClient.isAppointmentsRolledOutAt(prisonCode)) {
       true -> activitiesAppointmentsClient.getScheduledAppointments(prisonCode, date, nomisToDpsLocations.keys)
+        // Do not consider cancelled appointments
+        .filterNot { it.isCancelled }
```
matt-crowson-moj, line 59: You could avoid the comment and use an extension function if you fancied, meh

`private fun List<AppointmentSearchResult>.doNotConsiderCancelledAppointments() = filterNot { it.isCancelled }`

or something to this effect.

matt-crowson-moj (approved):
LGTM

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/locations/availability/BookedLocationsService.kt
timharrison-moj: It's fairly self-explanatory anyway - I'll maybe take the comment out. Next pr..

PR #473, opened 2025-04-25
BAVL-778: Availability check to take account of cancelled appointments via Prison API

timharrison-moj:
This PR:
* Retrieves and extra eventStatus field from Prison API for appointments at locations
* Ignores appointments that are cancelled (eventStatus=CANC) in the list
* Does not block rooms where cancelled appointments exist


matt-crowson-moj (approved):
LGTM

PR #474, opened 2025-04-25
BAVL-725 resolves an issue where the CVP link was provided for PRE/POST hearings when it should actually be the room link.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #475, opened 2025-04-25
BAVL-726 adding extra probation officer attributes to the schedule it em required for CSV extract.

matt-crowson-moj:
Needed for probation CSV extract [Jira](https://dsdmoj.atlassian.net/browse/BAVL-726).


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/resource/ScheduleResourceIntegrationTest.kt
```diff
@@ -28,6 +30,7 @@ import uk.gov.justice.digital.hmpps.hmppsbookavideolinkapi.service.events.Outbou
 import java.time.LocalDate
 import java.time.LocalTime
 
+@TestPropertySource(properties = ["feature.master.vlpm.types=true"])
```
matt-crowson-moj, line 33: Additional details (which are needed for this integration test) are only persisted when the master VLPM toggle is enabled.

Really we need to look at remove the toggle and the old code sitting behind the toggle.

PR #484, opened 2025-05-20
BAVL-849 adding a simple feature toggle to switch the mastering of public private notes on/off.

matt-crowson-moj:
I have kept this change fairly small with regards to impact on code footprint.

I expect we will also have toggles in the various UI's as well anyway.

### DIFF
--- a/helm_deploy/hmpps-book-a-video-link-api/values.yaml
+++ b/helm_deploy/hmpps-book-a-video-link-api/values.yaml
@@ -68,6 +68,7 @@ generic-service:
     sqs-domain-event-dlq-secret:
       HMPPS_SQS_QUEUES_BVLS_DLQ_NAME: "sqs_queue_name"
     feature-toggles:
+      FEATURE_MASTER_PUBLIC_PRIVATE_NOTES: "FEATURE_MASTER_PUBLIC_PRIVATE_NOTES"
 
   allowlist:
     groups:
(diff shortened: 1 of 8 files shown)


helm_deploy/hmpps-book-a-video-link-api/values.yaml
```diff
@@ -68,6 +68,7 @@ generic-service:
     sqs-domain-event-dlq-secret:
       HMPPS_SQS_QUEUES_BVLS_DLQ_NAME: "sqs_queue_name"
     feature-toggles:
+      FEATURE_MASTER_PUBLIC_PRIVATE_NOTES: "FEATURE_MASTER_PUBLIC_PRIVATE_NOTES"
```
matt-crowson-moj, line 71: The UI can use this same toggle (name).  So no need to add one by a different name.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/events/ActivitiesAndAppointmentsService.kt
```diff
@@ -36,7 +38,7 @@ class ActivitiesAndAppointmentsService(
       startTime = appointment.startTime,
       endTime = appointment.endTime,
       internalLocationId = nomisMappingService.getNomisLocationId(appointment.prisonLocationId)!!,
-      comments = appointment.comments,
+      comments = if (toggles.isMasterPublicAndPrivateNotes()) appointment.notesForPrisoners else appointment.comments,
       appointmentType = supportedAppointmentTypes.typeOf(appointment.bookingType()),
```
timharrison-moj, line 42: Just thinking through this, particularly around the time the toggle is set.  An existing booking might have comments, we switch the toggle, and the next time its updated we no longer created those comments on the appointment, instead we populate it with (potentially empty) notesForPrisoners. Might be right .. but it feels like there might be a middle bit of if (notesForPrisoners is empty and comments is not empty) still use the comments? Something like that..

timharrison-moj (approved):
LGTM.
Just thinking on the question raised.

PR #485, opened 2025-05-21
BAVL-849 ensure existing comments/notes are maintained when toggle is on/off.

matt-crowson-moj:
This change is to ensure existing comments or notes are not accidentally lost when the feature toggle is switch on/off.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/helper/ModelFactory.kt
```diff
@@ -334,7 +334,6 @@ fun requestProbationVideoLinkRequest(
 }
 
 fun amendCourtBookingRequest(
-  courtCode: String = "DRBYMC",
```
matt-crowson-moj, line 337: This was a redundant/unused test factory param, you cannot amend a court once a booking is created.

timharrison-moj (approved):
LGTM

PR #497, opened 2025-06-11
BAVL-925 initial Domain work to introduce new attributes on video booking for HTMCS number and guest pin. Service layer changes to come next.

matt-crowson-moj:
This is the first PR of a few to add support for new fields for court bookings only, HMTCS number and guest PIN.

These changes are in the Domain layer only.  Next up will be the service and model layers.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/VideoBooking.kt
```diff
@@ -196,6 +226,7 @@ class VideoBooking private constructor(
 
   companion object {
 
+    @Deprecated(message = "Use new court booking with CVP link details instead")
```
matt-crowson-moj, line 229: The deprecated methods will be removed in the next PR.

PR #498, opened 2025-06-12
BAVL-925 pulling the use of the new attributes into the service layer, telemetry and court reminder job.

matt-crowson-moj:
This brings the new attributes up to the service layer though still not fully in use.

It also addresses the court reminder job for missing url which now also needs to check the presence of the hmcts number and brings the telemetry events inline as well.


sameer-moj (approved):
LGTM

PR #504, opened 2025-06-20
BAVL-925 with the introduction of the HMCTS number we need to show this in the court CSV if this is provided instead of a video URL.

matt-crowson-moj:
We need to show the HMCTS number in the CVP link field in the court CSV if it is provided (instead of a video URL)


timharrison-moj (approved):
LGTM

PR #507, opened 2025-06-26
BAVL-951: Update to latest gradle spring boot plugin and add compiler options

timharrison-moj:
Compiler options:
* With latest spring boot - `-Xannotation-default-target=param-property` - to maintain default behaviour and suppress warnings.
* With LocalDate, LocalDateTime, LocalTime - `-Xwarning-level=IDENTITY_SENSITIVE_OPERATIONS_WITH_VALUE_TYPE:disabled` to suppress new warnings


build.gradle.kts
```diff
@@ -79,6 +79,8 @@ tasks {
   withType<KotlinCompile> {
     dependsOn("buildLocationsInsidePrisonApiModel", "buildActivitiesAppointmentsApiModel", "buildPrisonApiModel", "buildManageUsersApiModel")
     compilerOptions.jvmTarget = org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_21
+    compilerOptions.freeCompilerArgs.add("-Xannotation-default-target=param-property")
+    compilerOptions.freeCompilerArgs.add("-Xwarning-level=IDENTITY_SENSITIVE_OPERATIONS_WITH_VALUE_TYPE:disabled")
```
matt-crowson-moj, line 83: Just curious.  What does this actually mean/do?

matt-crowson-moj (approved):
LGTM

build.gradle.kts
matt-crowson-moj: When you say we need to amend those places, what is the actual change required then?

timharrison-moj: https://youtrack.jetbrains.com/issue/KT-78546/False-positive-IDENTITYSENSITIVEOPERATIONSWITHVALUETYPE-when-comparing-LocalDate-with-equality-operator

Effectively, avoid using `==` for dates, times, datetimes,  and use `.equals` instead.

timharrison-moj: But looks like if/when we do, IntelliJ might warn on it..

PR #508, opened 2025-06-27
BAVL-915 adding the daily schedule to the list of deployments that needs restarting when the FEATURE_HMCTS_LINK_GUEST_PIN is changed.

matt-crowson-moj:
The daily schedule service also needs restarting when the FEATURE_HMCTS_LINK_GUEST_PIN is changed.  It does not need its own file, it can piggy back on existing disable and enable scripts because the toggle name is shared across services.


timharrison-moj (approved):
LGTM

PR #509, opened 2025-07-01
BAVL-957: Adding pick up times feature scripts

timharrison-moj:


utility/disable-bvls-pick-up-times.sh
```diff
@@ -0,0 +1,51 @@
...
+# Apply the secret
+kubectl apply -f ./$SECRETS_FILE
+echo
+echo "Applied secrets to $NAMESPACE"
+
+rm -f ./$SECRETS_FILE
+
+# Restart the deployment pods
+echo
+echo "Restarting BVLS UI deployment on namespace $NAMESPACE"
```
matt-crowson-moj, line 49: restarting daily schedule?

timharrison-moj: Yes - this feature switch only affects the DS, nothing else.

timharrison-moj: Updated now, thanks.

matt-crowson-moj (approved):
LGTM

PR #510, opened 2025-07-02
BAVL-959 disable preprod jobs from running over the weekend due to dependent services turned off at weekends.

matt-crowson-moj:
The jobs fail to run at the weekend due to dependent DPS service downtime.


timharrison-moj (approved):
LGTM

PR #515, opened 2025-07-10
BAVL-914 this change removes references in the code to the old comments fields. The fields on the DB tables remains for now.

matt-crowson-moj:
This removes the references to the DB columns in the code.

Still to to:

- Run SQL to copy old comments data to 'notes for staff' columns.
- Another PR which drops the comments columns from the tables.


utility/sql/copy-comments-to-notes.sql
matt-crowson-moj, line 1: The will work as is but the dates will need changing accordingly when running.

matt-crowson-moj, line 1: These will work as is but the dates will need changing accordingly when running.  Included so can see what would be run (and verify SQL is correct).

PR #516, opened 2025-07-11
BAVL-949 it is now safe to remove the comments field as all data has been migrated accross to the notes_for_staff field.

matt-crowson-moj:
It is now safe to remove this field as it is no longer used inside or outside the API.

It has been superseded by notes_for_staff.

All old comments data has been moved across to the notes_for_staff in production.


sameer-moj (approved):
LGTM

PR #527, opened 2025-07-29
BAVL-1001 bug fix

matt-crowson-moj:
This change fixes a bug whereby prisons with no new rooms were also being included in the email that is sent.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/emails/administration/AdministrationEmailServiceTest.kt
```diff
@@ -162,12 +162,12 @@ class AdministrationEmailServiceTest {
 
   @Test
   fun `should send no emails when no new rooms`() {
-    whenever(prisonsService.getListOfPrisons(true)) doReturn emptyList()
```
matt-crowson-moj, line 165: The test fixture set up was not quite right therefore the expected assertions were also not quite right.

PR #533, opened 2025-08-06
BAVL-1001 add cron for running new rooms job - at 09:00 on every day-of-week from Monday through Friday.

matt-crowson-moj:
Run the new rooms job at 9AM from Monday through Friday.


timharrison-moj (approved):
LGTM

PR #534, opened 2025-08-06
BAVL-1024 tidy up validation errors for UI purposes when reporting.

matt-crowson-moj:
The UI shows the user error and in this situation it currently shows the following.  
image

This change removes the `Validation failure` prefix.


timharrison-moj (approved):
LGTM

PR #535, opened 2025-08-06
BAVL-1026 changing create appointment call to A&A to use DPS location ID instead of the NOMIS location ID.

matt-crowson-moj:
The purpose of this change is move towards using DPS location IDs instead of NOMIS locations IDs when creating appointments via the A&A API (now that A&A supports DPS location IDs).

This only changes the creation of appointments at the moment.

I am doing the changes in stages to keep the changes small.  Once in dev, I will test and make sure service still works as expected.


timharrison-moj (approved):
LGTM

PR #538, opened 2025-08-07
BAVL-1026 changing search appointment call to A&A to use DPS location ID instead of the NOMIS location ID.

matt-crowson-moj:
The is last change (search calls) in moving to using the DPS location ID when call the A&A API.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/activitiesappointments/ActivitiesAppointmentsClient.kt
```diff
@@ -73,29 +73,25 @@ class ActivitiesAppointmentsClient(private val activitiesAppointmentsApiWebClien
...
-    getPrisonersAppointments(prisonCode, prisonerNumber, onDate)
-      .filter { locationIds.toList().contains(it.internalLocation?.id) }
-  } else {
-    emptyList()
+  fun getPrisonersAppointments(prisonCode: String, prisonerNumber: String, onDate: LocalDate, locationId: UUID) = run {
+    log.info("A&A CLIENT: query params - prisonCode=$prisonCode, prisonerNumber=$prisonerNumber, onDate=$onDate, locationId=$locationId")
+    activitiesAppointmentsApiWebClient.post()
+      .uri("/appointments/{prisonCode}/search", prisonCode)
+      .bodyValue(
+        AppointmentSearchRequest(
```
matt-crowson-moj, line 83: So this request does also support category code (appointment type) we might be able to harness this as well.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/activitiesappointments/ActivitiesAppointmentsClientTest.kt
```diff
@@ -50,47 +47,6 @@ class ActivitiesAppointmentsClientTest {
...
-  }
-
-  @Test
-  fun `should get multiple prisoner appointments at locations`() {
-    server.stubGetPrisonersAppointments(
-      prisonCode = BIRMINGHAM,
-      prisonerNumber = "123456",
-      date = tomorrow(),
-      locationIds = setOf(1000, 2000, 3000),
-    )
```
timharrison-moj, line 74: This would suggest a usage for multiple locations?

matt-crowson-moj: Yes but it was never used that way.

PR #539, opened 2025-08-08
BAVL-1031 adding timezone config to reminder jobs so they run at correct Europe/London time and not UTC.

matt-crowson-moj:


timharrison-moj (approved):
LGTM

PR #544, opened 2025-09-01
BAVL-1058: Update gradle springboot plugin to v9.0.0 and GH workflows

timharrison-moj:
The aim is to get the overnight Github actions security scans working with the v2 NVD feed. 
The DEV CP namespace has had its GH terraform integrations updated from v5.39.0 to v6.6.0.


iterative-it (approved):
LGTM

PR #558, opened 2025-10-03
BAVL-1101 changes for supporting temporary blocking of a rooms.

matt-crowson-moj:
This change introduces a new status for decorating rooms - TEMPORARILY_BLOCKED.

This differs to INACTIVE rooms, with INACTIVE the room can never be booked.  With TEMPORARILY_BLOCKED it also includes a date range (from and to) to indicate when the room is temporarily not available.

Any checks for availability will exclude any room where the date of the booking overlaps with the blocked date range.

I have tested running this locally with the BVLS UI to make sure the changes are backwards compatible, this includes creation and amending of both court and probation bookings.  The removed attributes are not actually used in the UI therefore their removal has no impact.

No feature toggle is required for this change due to it being backwards compatible with the existing UI.

Still to do in separate tasks:

- Audit changes to the status using App Insights telemetry.
- Job to reset temporarily blocked rooms to `active` then the `to date` has passed.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
@@ -34,8 +34,6 @@ class LocationAttribute private constructor(
   @JoinColumn(name = "prison_id")
   val prison: Prison,
 
-  val statusMessage: String? = null,
-
   val createdBy: String,
 
   val createdTime: LocalDateTime = LocalDateTime.now(),
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/model/Location.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/model/Location.kt
@@ -37,9 +37,6 @@ data class RoomAttributes(
   @Schema(description = "The status of the room (ACTIVE, INACTIVE, TEMPORARILY_BLOCKED)", example = "ACTIVE", required = true)
   val locationStatus: LocationStatus,
 
-  @Schema(description = "An optional message relating to an inactive status", example = "Room damaged")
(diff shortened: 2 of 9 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
```diff
@@ -144,11 +129,18 @@ class LocationAttribute private constructor(
...
-  fun isAvailableFor(probationTeam: ProbationTeam, onDate: LocalDate, startTime: LocalTime, endTime: LocalTime): AvailabilityStatus = check(probationTeam, onDate, startTime, endTime)
-
-  private fun check(probationTeam: ProbationTeam, onDate: LocalDate, startTime: LocalTime, endTime: LocalTime): AvailabilityStatus {
+  fun isAvailableFor(usageType: LocationUsageType, onDate: LocalDate, startTime: LocalTime, endTime: LocalTime): AvailabilityStatus = run {
     if (locationStatus == LocationStatus.INACTIVE) return AvailabilityStatus.NONE
 
+    if (locationStatus == LocationStatus.TEMPORARILY_BLOCKED && onDate.between(blockedFrom!!, blockedTo)) return AvailabilityStatus.NONE
+
+    when (usageType) {
+      is Court -> check(usageType, onDate, startTime, endTime)
```
timharrison-moj, line 138: Should there be a Shared usage type here? Or is usageType the type of booking?

timharrison-moj (approved):
LGTM. One small question, but I think that's covered within each check() function for court and probation.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
matt-crowson-moj: It's just a naming thing.  I'll try and think of a better name 🤔

PR #560, opened 2025-10-07
BAVL-1103 changes to introduce a new telemetry event for location attribute status changes for auditing purposes.

matt-crowson-moj:


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/telemetry/LocationAttributeTelemetryEvent.kt
```diff
@@ -0,0 +1,41 @@
...
+      }
+    } else {
+      require(attribute.amendedBy == raisedBy.username) {
+        "Cannot create location status telemetry event, amended by user does not match raised by user."
+      }
+    }
+  }
+
+  override fun properties(): Map<String, String> = run {
+    buildMap {
```
matt-crowson-moj, line 23: I think this covers the main things.  Let me know if you think differently.

timharrison-moj: Maybe usage?  ie.. is it Court, Probation, Shared, or Schedule, but everything else fine

timharrison-moj (approved):
LGTM - one suggestion, but good as is.

PR #621, opened 2026-03-27
Revert "BAVL-1201 correct JPA annotations on entities which should be many to one and not one to one."

matt-crowson-moj:
Reverts ministryofjustice/hmpps-book-a-video-link-api#596


timharrison-moj (approved):
LGTM

PR #633, opened 2026-04-15
BAVL-1295 removing old email code now the refactored email code is active in production.

matt-crowson-moj:
Now that the new email code is active in production it is now safe to remove the old email code and feature toggle hiding it.


timharrison-moj (approved):
LGTM

PR #634, opened 2026-04-16
BAVL-1297 adding shell script to making change the Gov Notify email key simpler.

matt-crowson-moj:
There was no script for doing this so added one as less likely to make errors.

This has been used for dev and preprod.


timharrison-moj (approved):
LGTM

PR #635, opened 2026-04-17
BAVL-1251 refactoring in preparation for reschedule booking emails work, no functional changes.

matt-crowson-moj:
The changes in this PR are simply repackaging/moving files in preparation for upcoming rescheduling emails work.

There is now a dedicated facade package which is essentially the root of all other changes.

There are no actual functional changes.


timharrison-moj (approved):
LGTM

PR #638, opened 2026-04-24
BAVL-1269:  Default undecorated rooms to not be available for bookings.

timharrison-moj:
We might need to decorate more rooms in DEV..


matt-crowson-moj (approved):
LGTM

PR #639, opened 2026-04-27
BAVL-1269 undecorated rooms are no longer available by default. Defaulting to out of use for CSV download.

matt-crowson-moj:
This is part 2 on the back of this change to room availability for undecorated rooms [PR](https://github.com/ministryofjustice/hmpps-book-a-video-link-api/pull/638).

Rooms that are undecorated no longer default to shared availability.  Going forwards they are not available.  This change needs to be reflected this on the CSV download also by saying the room is 'Out of use' until the room is decorated.


timharrison-moj (approved):
LGTM

PR #643, opened 2026-04-29
BAVL-1194 trial increasing POD memory in preprod prior to chaning production.

matt-crowson-moj:
Attempt to change preprod memory settings first to confirm changes take effect before trying in production.

### DIFF
--- a/helm_deploy/values-preprod.yaml
+++ b/helm_deploy/values-preprod.yaml
@@ -23,6 +23,14 @@ generic-service:
     groups:
       - internal
 
+  resources:
+    requests:
+      cpu: 10m
+      memory: 100Mi
+    limits:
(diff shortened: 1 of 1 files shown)


helm_deploy/values-preprod.yaml
```diff
@@ -23,6 +23,14 @@ generic-service:
     groups:
       - internal
 
+  resources:
+    requests:
+      cpu: 10m
+      memory: 100Mi
+    limits:
```
matt-crowson-moj, line 30: Was advised to double the limits and leave requests the same.  Limits cpu and memory are double that of the cloud platform configuration.

timharrison-moj (approved):
LGTM

PR #644, opened 2026-04-29
BAVL-1194 increase POD memory in prod to help alleviate out of memory issues.

matt-crowson-moj:
Increase pod memory to help with **out of memory** issues we are seeing in production.

Looking at the memory stats were were maxing out, advice from the SRE team was to double what we had.

**Note: we do have other stuff that would be released if this were to go out now.**


timharrison-moj (approved):
LGTM

PR #645, opened 2026-04-29
BAVL-1296: Saving the SQL for checking/setting probation team flags

timharrison-moj:
Just storing the useful SQL for checking and setting the probation team flags.


matt-crowson-moj (approved):
LGTM

PR #648, opened 2026-05-08
BAVL-1251 method rename to aid intent/meaning.

matt-crowson-moj:
Method rename only, non-functional change.


timharrison-moj (approved):
LGTM

PR #649, opened 2026-05-11
BAVL-1329 add rollout script to assist with rolling out changes to secrets.

matt-crowson-moj:
Add support to manage k8s secrets via one rollout control file.


timharrison-moj (approved):
LGTM.

PR #650, opened 2026-05-13
BAVL-1291: New social security tribunal courts added for the North West

timharrison-moj:
This adds 11 new courts for the North West.
Contacts will be added manually to preprod and prod as a follow-on action.


matt-crowson-moj (approved):
LGTM

PR #651, opened 2026-05-13
BAVL-1326 migrate security jobs to bring inline with latest template security changes.

matt-crowson-moj:
Update the repo security actions to bring inline with the latest template changes.


timharrison-moj (approved):
LGTM

PR #652, opened 2026-05-13
BAVL-1334 security patch API.

matt-crowson-moj:
Trivy ignore file replaced with Snyk ignore file inline with the latest gradle spring boot plugin and also this change [here](https://github.com/ministryofjustice/hmpps-book-a-video-link-api/pull/651)


timharrison-moj (approved):
LGTM

PR #653, opened 2026-05-20
BAVL-1346 enable rescheduled emails feature in production.

matt-crowson-moj:
The changes turns on the rescheduled emails feature on in production.

Once live (for a while) we can remove the old redundant code and toggle.


timharrison-moj (approved):
LGTM

PR #655, opened 2026-05-21
BAVL-1347 include snyk file which was missed off previous patch commit.

matt-crowson-moj:
Missed off previous security patch commit.


timharrison-moj (approved):
LGTM.

PR #657, opened 2026-06-05
BAVL-1376 change Snyk workflow to use auth credentials instead of API key.

matt-crowson-moj:
Use latest approach for Snyk security workflow.  This brings it inline with recommendations from the SRE team and kotlin template project.


timharrison-moj (approved):
LGTM

PR #660, opened 2026-06-15
BVLS-1345 resolve issue where exisiting slot not included for timeslot service availability checks.

matt-crowson-moj:
This change should resolve an issue whereby existing probation bookings during availability checking were being excluded.

It is similar in some ways to the date time checking service however we only have a date and slot to work with from the amend request.  To work with this we check the slot is the same and then the existing booking is also the same, if so we include it.


timharrison-moj (approved):
LGTM

PR #669, opened 2026-06-23
BAVL-1309 refactoring the models to remove all notnull annotations due to swagger changes, in preparation for location ID changes.

matt-crowson-moj:
This is a refactoring PR to remove all NotNull annotations due to swagger API generations changes now being based off the type and not the annotation.  As a result fields that had the NotNull annotation but were previously nullable are no longer nullable.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/integration/SqsIntegrationTestBase.kt
```diff
@@ -33,7 +33,7 @@ abstract class SqsIntegrationTestBase : IntegrationTestBase() {
 
   protected fun waitUntil(fn: () -> Unit) {
     // Default wait is 10 seconds, so dropping down to 5.
-    await.atMost(5, TimeUnit.SECONDS) untilAsserted {
+    await.atMost(10, TimeUnit.SECONDS) untilAsserted {
```
matt-crowson-moj, line 36: Started to see an intermittent failure on one integration test. I couldn't see any obvious issues with the test so increased to see if this resolves it.

timharrison-moj (approved):
LGTM.
Probably needs a full sweep through testing session on DEV?

PR #671, opened 2026-06-25
BAVL-1309 reworking the search endpoint to support search by both location ID or location key.

matt-crowson-moj:
This change prepares the search endpoint predominantly used by external UI services e.g. A&A to allow searching by location ID instead of location key.

When all calling services are migrated to using the ID then the key can be removed from the API endpoint.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/VideoLinkBookingsService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/VideoLinkBookingsService.kt
@@ -58,22 +58,27 @@ class VideoLinkBookingsService(
   }
 
   fun findMatchingVideoLinkBooking(searchRequest: VideoBookingSearchRequest, user: User): VideoLinkBooking {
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/VideoLinkBookingsService.kt
```diff
@@ -58,22 +58,27 @@ class VideoLinkBookingsService(
   }
 
   fun findMatchingVideoLinkBooking(searchRequest: VideoBookingSearchRequest, user: User): VideoLinkBooking {
-    val location = locationsService.getLocationByKey(searchRequest.locationKey)
-      ?: throw EntityNotFoundException("Location with key ${searchRequest.locationKey} not found")
+    // TODO when UI's moved over to location ID can remove the location key.
+    val locationId = when {
+      searchRequest.dpsLocationId != null -> locationsService.getLocationById(searchRequest.dpsLocationId)?.dpsLocationId
```
matt-crowson-moj, line 63: Possibly overkill searching by ID here given we are provided an ID on the request.

timharrison-moj (approved):
LGTM

PR #672, opened 2026-06-26
BAVl-1390: New user type for nDelius probation users

timharrison-moj:
Updated all endpoints, facades, services and tests to verify probation access for nDelius users.
(they are checked in the same was as ExternalUser with `isProbation = true`).


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/facade/BookingFacade.kt
```diff
@@ -99,34 +103,34 @@ class BookingFacade(
...
-
-  fun cancel(videoBookingId: Long, cancelledBy: ServiceUser) {
-    // No events should be raised on the back of calling this function, this is by design.
-    val booking = videoBookingServiceDelegate.cancel(videoBookingId, cancelledBy)
-    log.info("Video booking ${booking.videoBookingId} cancelled by user type ${cancelledBy::class.simpleName}")
     emailFacade.sendEmails(BookingAction.CANCEL, booking, getPrisoner(booking.prisoner()), cancelledBy)
     trackTelemetry(BookingAction.CANCEL, booking, cancelledBy)
   }
 
+  // Administrative action by service user
```
matt-crowson-moj, line 133: Not now but I think perhaps we could pull the admin functions into their own facade ...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/UserService.kt
```diff
@@ -66,6 +79,25 @@ class UserService(
...
+      AuthSource.delius -> {
+        /**
+         * The user account is held in nDelius with their access controlled by roles allocated there.
+         * Roles in nDelius are mapped to DPS roles by HMPPS Auth / manage-users-api at login.
+         * Group memberships are not relevant to nDelius users.
+         * The usernames are not email addresses.
+         * Delius users are always probation users.
+         */
+
+        DeliusUser(
```
matt-crowson-moj, line 91: I thought the plan was to reduce it down to one User class?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/telemetry/ProbationBookingCancelledTelemetryEvent.kt
```diff
@@ -29,16 +30,16 @@ class ProbationBookingCancelledTelemetryEvent private constructor(
 
   companion object {
     fun user(booking: VideoBooking, user: User): ProbationBookingCancelledTelemetryEvent {
-      require(user is PrisonUser || user is ServiceUser || (user is ExternalUser && user.isProbationUser)) {
-        "Can only create probation cancelled metric for service, prison or court users."
+      require(user is PrisonUser || user is ServiceUser || user is DeliusUser || (user is ExternalUser && user.isProbationUser)) {
```
matt-crowson-moj, line 36: Not here but I have a general question on telemetry capture.  In the probation telemetry we have:

```
override fun properties() = booking.commonProperties().plus("created_by" to if (booking.createdByPrison) "prison" else "probation")
```

Does this need to change?  Do these need to cater for Delius users?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/UserService.kt
timharrison-moj: Yes, I pulled away from that when I saw the number of places affected and the rework it would require. I think where we get another user source eg. justice domain, or a solicitor auth method, we could reassess.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/telemetry/ProbationBookingCancelledTelemetryEvent.kt
timharrison-moj: Yes - I'll make a small change here to record whether external or delius probation users.

matt-crowson-moj: Just checking but there is more than one instance of this logic i.e. not just in the cancel.

timharrison-moj: Updated all of create, cancel and amend telemetry events to set a new property of delius_user=true or false. Might come in handy to know how many probation users are switching to their ndelius accounts later.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/ChangeTrackingService.kt
```diff
@@ -29,7 +29,7 @@ class ChangeTrackingService(
    * Will return the change type NONE if there are no actual changes.
    */
   fun determineChangeType(videoBookingId: Long, requestedBookingChanges: AmendVideoBookingRequest, amendedBy: User): ChangeType {
-    require(amendedBy is PrisonUser || amendedBy is ExternalUser) {
+    require(amendedBy is PrisonUser || amendedBy is ExternalUser || amendedBy is DeliusUser) {
       "Only prison users and external users are supported. ${amendedBy::class.simpleName} is not supported."
```
matt-crowson-moj: Is this error message missing the Delius user type?

matt-crowson-moj (dismissed):
LGTM - one small question on error message added since last review.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/ChangeTrackingService.kt
timharrison-moj: Yes. Will update.

timharrison-moj: Done now.

matt-crowson-moj (approved):
LGTM

PR #673, opened 2026-06-29
BAVL-1384 can only merge future meetings that are active. Cancelled bookings cannot be amended.

matt-crowson-moj:
You can only merge active probation bookings via the booking history service.  You cannot amend inactive (cancelled) bookings.


timharrison-moj (approved):
LGTM

PR #674, opened 2026-06-29
BAVL-1309 API will now support both create and amend bookings using either location IDs or location keys.

matt-crowson-moj:
These changes mean it should now be safe to change the UI's creating and amending bookings to be updated to use location IDs instead of keys.

To do this the UI's must be using ID's in the journey objects.  This will likely mean change any place keys are used (e.g. location drop-down/selects will also need to be using ID's which are already available from the API location lookups endpoints).


timharrison-moj (approved):
LGTM

PR #675, opened 2026-06-30
BAVL-1309 missed a location ID versus key in the ChangeTrackingService.

matt-crowson-moj:
Testing UI changes locally highlighted the issue when I was supply only a DPS location ID.  The ChangeTrackingService was still set up to only work with location keys.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/ChangeTrackingService.kt
```diff
@@ -124,6 +125,8 @@ private fun AmendVideoBookingRequest.toComparableBooking(
   )
 }
 
+private fun Appointment.location(locationsService: LocationsService) = dpsLocationId ?: locationsService.getLocationByKey(locationKey!!)?.dpsLocationId
```
matt-crowson-moj, line 128: A lot of these lookups can go when we have switch the UI's over to IDs.

timharrison-moj (approved):
LGTM

PR #678, opened 2026-07-09
BAVL-1399 add blocking from and to times to the request and response model objects to help define the initial contract.

matt-crowson-moj:
This PR adds the blocking from and to times to aid define initial contract for any callers.

Note the values are not persisted at this point.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/locationsinsideprison/LocationsInsidePrisonClient.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/locationsinsideprison/LocationsInsidePrisonClient.kt
@@ -36,6 +36,7 @@ class LocationsInsidePrisonClient(private val locationsInsidePrisonApiWebClient:
     .onErrorResume(WebClientResponseException.NotFound::class.java) { Mono.empty() }
     .block()
 
+  @Deprecated("Use getLocationsById instead")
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/entity/LocationAttribute.kt
```diff
@@ -150,7 +150,6 @@ class LocationAttribute private constructor(
     when (locationUsage) {
       LocationUsage.SHARED -> AvailabilityStatus.SHARED
       LocationUsage.PROBATION -> when {
-        // TODO need to add check here for the court or sentence attribute
```
matt-crowson-moj, line 153: This TODO should not be here.  This was previously added when we though the checking of these would be done on a non-schedule decoration.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/client/locationsinsideprison/LocationsInsidePrisonClient.kt
```diff
@@ -36,6 +36,7 @@ class LocationsInsidePrisonClient(private val locationsInsidePrisonApiWebClient:
     .onErrorResume(WebClientResponseException.NotFound::class.java) { Mono.empty() }
     .block()
 
+  @Deprecated("Use getLocationsById instead")
```
matt-crowson-moj, line 39: Added whilst I am here to act as a reminder.

timharrison-moj (approved):
LGTM.

PR #682, opened 2026-07-13
BAVL-1399 validate blocked from time and to time are valid.

matt-crowson-moj:
Make sure we are given valid blocked from and to times for blocked rooms for both create and amend i.e. from is before to.


timharrison-moj (approved):
LGTM

PR #684, opened 2026-07-15
BAVL-1400: Add feature switch for room blocking times into the rollout control script

timharrison-moj:
Script changes only.


matt-crowson-moj (approved):
LGTM

PR #685, opened 2026-07-16
BAVL-1410 update location API types to support new VIDEO_LINK service type.

matt-crowson-moj:
This change pre-empts potential deserialisation issues in prod.

A new service type has been added to locations, without an up to data spec it can cause deserialisation issues when looking up locations from the locations API.


timharrison-moj (approved):
LGTM

PR #691, opened 2026-07-29
BAVL-1361: Added feature switches for availability checker views

timharrison-moj:
And added both the weekly view and timeline view secrets to dev, pre and prod k8s feature-toggle secrets.
All set to false for now - can be toggled using the script.


matt-crowson-moj (approved):
LGTM

PR #693, opened 2026-08-03
BAVL-1425: Adding south central probation teams

timharrison-moj:
Adds 18 more probation teams - these are court teams.
Contacts will be added manually in PREPROD and PROD (there is a script to generate these, but they are applied manually so contact details are not in GH).


matt-crowson-moj (approved):
LGTM

PR #694, opened 2026-08-04
BAVL-1431: Removed prisoner number from each report section.

timharrison-moj:
The prisoner number is provided to each SAR report via the PRN parameter.

OSAR team requested it not also be printed for each report line.
Also noticed README had a slight mistake - wrong test name to generated SAR files.
Updated the SAR HTML file to compare against.


matt-crowson-moj (approved):
LGTM

PR #695, opened 2026-08-06
BAVL-1428: Gradle plugin update

timharrison-moj:
Also updates swagger-ui to latest version and removes a constraint from the build.gradle.kts file.


matt-crowson-moj (approved):
LGTM

PR #696, opened 2026-08-12
BAVL-1426 show all A&A appointments booked into video link rooms instead of just A&A video type appointments.

matt-crowson-moj:
Prior to this change we would only see video type appointments from A&A being included as part of the video event results.  

This would not give a true representation of room usage because any A&A appointment can book into a video based room.

With this change we will now see all A&A appointments apart from the BVLS specific types (COURT and PROBATION) as these are mastered in BVLS (but synced to A&A so can be ignored).

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/VideoEventsByLocationService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/VideoEventsByLocationService.kt
@@ -35,8 +35,6 @@ class VideoEventsByLocationService(
     private val log = LoggerFactory.getLogger(this::class.java)
   }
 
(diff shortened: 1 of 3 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsbookavideolinkapi/service/VideoEventsByLocationService.kt
```diff
@@ -86,6 +84,8 @@ class VideoEventsByLocationService(
     )
   }
 
+  private fun List<AppointmentSearchResult>.filterOutBvlsAppointmentTypes() = filterNot { listOf("VLB", "VLPM").contains(it.appointmentCode()) }
```
matt-crowson-moj, line 87: VLB = court
VLPM = probation

prabash-moj (approved):

